PR #3, opened 2022-04-25
feat: messaging

vsomayaji:
This adds messaging support (using [PubNub](https://www.pubnub.com/)) to the `PlatformClient`.

If the `PlatformClient` is configured with `PlatformMessagingKeys` (optional, since some clients may not support text-based communication), the client will be able to use two methods on a service request's `Room`: `messageStream` to stream all messages sent and received, and `sendMessage` to send a message.


example/lib/main.dart
```diff
@@ -78,108 +99,175 @@ class _MyAppState extends State<MyApp> implements RoomHandler {
...
+          child: const Text('End Call'),
+        ),
+      ],
+    );
+  }
+
+  Widget get _messagingWidget {
+    return Row(
+      children: <Widget>[
+        Expanded(
```
vsomayaji, line 147: Without `Expanded`, sticking `TextFormField` in a `Row` causes layout issues.

https://stackoverflow.com/a/56923281

lib/src/models/message.dart
```diff
@@ -0,0 +1,23 @@
+class Message {
+  /// Whether the message was sent by the Explorer.
+  final bool isLocal;
```
vsomayaji, line 3: I'm not a huge fan of this name. I toyed with names like `sentByExplorer`, `fromExplorer` and `fromMe`, but none of them felt great. For now, since we're using "local" and "remote" when referring to the Explorer and Agent streams, we'll use the same terms here.

PR #5, opened 2022-05-04
fix: service request is not cancelled when disconnecting ungracefully

vsomayaji:
If a service request is in the queue and the Explorer disconnects ungracefully (e.g. quits the app, closes the browser, or loses their network connection), the request will stay in the queue until an Agent picks it and flags it as a did-not-connect (DNC). This wastes Agent time and negatively impacts our DNC metrics.

With this PR, we use MQTT's [Last Will and Testament](https://www.hivemq.com/blog/mqtt-essentials-part-9-last-will-and-testament/) feature to cancel the queued service request when the MQTT client is ungracefully disconnected. This required changing the lifecycle of the MQTT client. Since the LWT message is configured when the MQTT client connects to the broker, instead of connecting after signing in, we're connecting after creating a service request.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -15,7 +15,6 @@ import 'models/service_request.dart';
 import 'models/session.dart';
 import 'models/track.dart';
 import 'platform_exceptions.dart';
-import 'platform_mq.dart';
 import 'room.dart';
 
 /// The Platform client.
@@ -30,7 +29,6 @@ class PlatformClient {
 
   final PlatformClientConfig _config;
   Session? _session;
-  PlatformMQ? _mq;
   pn.PubNub? _pubnub;
 
   int get _userId => _session!.userId;
@@ -48,9 +46,6 @@ class PlatformClient {
   /// After this is called, the object is not in a usable state and should be discarded.
   void dispose() {
     _httpClient.close();
-
-    _mq?.dispose();
-    _mq = null;
   }
 
   /// Sends a verification code to a phone number.
@@ -169,8 +164,6 @@ class PlatformClient {
   Future<void> logout() async {
     // TODO: Actually log out. For now, we're copying the legacy apps and just removing the token.
     _session = null;
-    _mq?.dispose();
-    _mq = null;
   }
 
   /// Creates a service request for the logged-in user.
@@ -194,7 +187,7 @@ class PlatformClient {
     ServiceRequest serviceRequest =
         ServiceRequest.fromJson(await _httpPost('/api/user/$_userId/service-request', body));
 
-    return KurentoRoom.create(_config.environment, this, _mq!, _pubnub, serviceRequest, roomHandler);
+    return KurentoRoom.create(_config.environment, this, _session!, _pubnub, serviceRequest, roomHandler);
   }
 
   /// Cancels a service request.
@@ -397,8 +390,6 @@ class PlatformClient {
       return body;
     } else if (body['response']?['errorCode'] == 'SEC-001') {
       _session = null;
-      _mq?.dispose();
-      _mq = null;
       throw const PlatformInvalidTokenException();
     } else if (body['response']?['errorMessage'] != null) {
       throw PlatformLocalizedException(body['response']?['errorCode'], body['response']['errorMessage']);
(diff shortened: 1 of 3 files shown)


lib/src/platform_client.dart
```diff
@@ -408,10 +399,6 @@ class PlatformClient {
   }
 
   void _initSession() {
-    // Initialize the RabbitMQ client.
-    _mq?.dispose();
-    _mq = PlatformMQImpl(_config.environment, _session!);
```
vsomayaji, line 413: Before, we were connecting to the MQTT broker after logging in and disconnecting when logging out. There were a couple issues with this approach:
1. Since we only need the MQTT connection during service requests, we could be wasting resources.
2. We don't have enough information after logging in to create a meaningful LWT message.

lib/src/platform_mq.dart
```diff
@@ -48,6 +44,14 @@ class PlatformMQImpl implements PlatformMQ {
       ..onSubscribeFail = _handleSubscribeFail
       ..onUnsubscribed = _handleUnsubscribed;
 
+    if (lastWillMessage != null && lastWillTopic != null) {
+      _client.connectionMessage = MqttConnectMessage()
+        ..startClean()
+        ..withWillMessage(lastWillMessage)
+        ..withWillQos(MqttQos.atMostOnce)
```
vsomayaji, line 51: For now, we're using at-most-once (QoS 0) everywhere.

lib/src/platform_mq.dart
```diff
@@ -85,7 +89,7 @@ class PlatformMQImpl implements PlatformMQ {
 
     Subscription? subscription = _client.subscribe(topic, qosLevel);
     if (subscription == null) {
-      _handleSubscribeFail(topic);
+      throw const PlatformUnknownException('Topic subscription failed');
```
vsomayaji, line 92: I spotted this :bug: while I was in here. `_handleSubscribeFail` just logs right now, so we may have been missing errors.

lib/src/room.dart
```diff
@@ -156,6 +159,17 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+  // If the MQTT client disconnects ungracefully, the last-will message will cancel the service request if it is still
+  // queued using Platform's (deprecated but functional) ServiceRequestsListener.
+  String get _lastWillMessage => jsonEncode({
+        'action': 'CANCEL',
+        'requestType': 'AIRA',
+        'serviceid': _serviceRequest.id,
+        'userid': _serviceRequest.userId,
+      });
+
+  String get _lastWillTopic => '${_env.name}/sr/req';
```
vsomayaji, line 171: The [`ServiceRequestListener`](https://github.com/aira/platform/blob/1ae4c68f2cb0d66ad5851c50765316b666f0ebb8/core/server/AiraPlatform/src/main/java/io/aira/amqp/ServiceRequestsListener.java#L135-L139) will only cancel the service request if it is queued (i.e. the status is `REQUEST`). If it is in a different state, it won't do anything.

lib/src/room.dart
```diff
@@ -242,8 +256,7 @@ class KurentoRoom extends ChangeNotifier implements Room {
 
     await _messageSubscription?.dispose();
 
-    await _mq.unsubscribe(_serviceRequestPresenceTopic);
-    await _mq.unsubscribe(_participantTopic);
+    _mq.dispose();
```
vsomayaji, line 259: We don't have to unsubscribe from the individual topics, since `dispose` disconnects from the broker and deletes all subscriptions.

lib/src/room.dart
```diff
@@ -282,6 +295,17 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+        if (serviceRequestState == ServiceRequestState.queued) {
+          // HACK: If the Agent is sending audio and we still think we're queued, we're not receiving messages on the
+          // service request presence topic. Until we can figure out why this is happening, pretend we received a
+          // message and transition the service request status to started.
+          _log.shout('missed service request status message topic=$_serviceRequestPresenceTopic');
+          _agentName = '';
+          _serviceRequestState = ServiceRequestState.started;
+          _updateParticipantStatus();
+          notifyListeners();
+        }
```
vsomayaji, line 308: I'm hoping that changing the lifecycle of the MQTT client will eliminate the failures to subscribe to the service request presence topic, but in case it doesn't, this hack is here as a failsafe. If it turns out we don't need it, we'll remove it.

PR #6, opened 2022-05-05
fix: unawaited futures

vsomayaji:


analysis_options.yaml
```diff
@@ -1,4 +1,29 @@
...
+  # https://dart-lang.github.io/linter/lints/index.html.
+  #
+  # Instead of disabling a lint rule for the entire project in the
+  # section below, it can also be suppressed for a single line of code
+  # or a specific dart file by using the `// ignore: name_of_lint` and
+  # `// ignore_for_file: name_of_lint` syntax on the line or in the file
+  # producing the lint.
+  rules:
+    prefer_single_quotes: true
+    unawaited_futures: true
```
vsomayaji, line 26: https://dart-lang.github.io/linter/lints/unawaited_futures.html

lib/src/platform_mq.dart
```diff
@@ -10,10 +10,10 @@ import 'platform_client.dart';
 import 'platform_exceptions.dart';
 import 'platform_mq_server_client.dart' if (dart.library.html) 'platform_mq_browser_client.dart' as mqttsetup;
 
-typedef MessageCallback = void Function(String message);
+typedef MessageCallback = Future<void> Function(String message);
```
vsomayaji, line 13: Changed the return type to `Future<void>` so we can catch and log exceptions.

lib/src/platform_mq.dart
```diff
@@ -10,10 +10,10 @@ import 'platform_client.dart';
...
 
 abstract class PlatformMQ {
-  Future<bool> get isConnected;
+  bool get isConnected;
```
vsomayaji, line 16: I can't recall why this was a `Future<bool>`. A `bool` makes much more sense.

PR #12, opened 2022-06-08
feat!: add `PlatformClient.createAccount`

vsomayaji:
BREAKING CHANGE: `PlatformClient.loginWithPhone` and `PlatformClient.loginWithEmail` have been removed. Now, once you have a phone or email verification code, you should call  `confirmPhoneVerificationCode` or `confirmEmailVerificationCode` to exchange the verification code for `Credentials`. Then, to create a `Session`, either `loginWithCredentials` (if `Credentials.isNewUser` is `false`) or `createAccount` (if `Credentials.isNewUser` is `true`).

### DIFF
--- a/lib/src/models/credentials.dart
+++ b/lib/src/models/credentials.dart
@@ -1,0 +1,8 @@
+class Credentials {
+  final String provider;
+  final String login;
+  final String password;
+  final bool isNewUser;
+
+  Credentials(this.provider, this.login, this.password, this.isNewUser);
+}
--- a/lib/src/models/profile.dart
+++ b/lib/src/models/profile.dart
@@ -5,3 +5,7 @@ enum Language {
   French,
   Spanish,
 }
+
+extension LanguageExtension on Language {
+  get name => toString().split('.').last;
+}
--- a/lib/src/models/profile.dart
+++ b/lib/src/models/profile.dart
@@ -1,0 +1,7 @@
+// ignore_for_file: constant_identifier_names
+
+enum Language {
+  English,
+  French,
+  Spanish,
+}
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -161,7 +161,7 @@ class PlatformClient {
     String body = jsonEncode({
       'authProvider': credentials.provider,
       'login': credentials.login,
-      'preferredLang': preferredLanguages,
+      'preferredLang': preferredLanguages?.map((language) => language.name),
       'tosAccepted': true,
       'verificationCode': credentials.password,
     });
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -10,6 +10,7 @@ import 'package:logging/logging.dart';
 import 'package:package_info_plus/package_info_plus.dart';
 import 'package:pubnub/pubnub.dart' as pn;
 
+import 'models/credentials.dart';
 import 'models/feedback.dart';
 import 'models/participant.dart';
 import 'models/service_request.dart';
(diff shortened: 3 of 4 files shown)


lib/src/models/credentials.dart
```diff
@@ -0,0 +1,8 @@
+class Credentials {
+  final String provider;
+  final String login;
+  final String password;
+  final bool isNewUser;
```
vsomayaji, line 5: I don't love `isNewUser` and am open to suggestions.

lib/src/models/profile.dart
```diff
@@ -0,0 +1,7 @@
+// ignore_for_file: constant_identifier_names
+
+enum Language {
```
vsomayaji, line 3: I stuck this in `profile.dart` as I anticipate that's where we'll stick other profile-related models. 

I created a `Language` enumeration since these values have to match the values supported by Platform.

ipainchaud: there is nothing wrong in putting Enums in their own files...

lib/src/platform_client.dart
```diff
@@ -62,6 +64,22 @@ class PlatformClient {
...
+  /// [loginWithCredentials].
+  Future<Credentials> confirmPhoneVerificationCode(String phoneNumber, String verificationCode) async {
+    String body = jsonEncode({
+      'authCode': verificationCode,
+      'phoneNumber': phoneNumber,
+    });
+
+    Map<String, dynamic> response = await _httpPost('/api/smartapp/verify/confirm', body);
+
+    return Credentials('PHONE_VERIFICATION', phoneNumber, response['verificationCode'], response['newUser']);
```
ipainchaud, line 80: For these const values like `PHONE_VERIFICATION`, I usually prefer to see them defined as a static const variable to avoid typos which your eyes have a hard time to see and for which the compiler could help...

In this case though, it would not be used almost at all... keep it as is.

lib/src/platform_client.dart
```diff
@@ -62,6 +64,22 @@ class PlatformClient {
...
+  /// [phoneNumber] is the phone number in E.164 format and [verificationCode] is the verification code. If successful,
+  /// the returned [Credentials] can be used to [createAccount] (if [Credentials.isNewUser] is `true`) or to
+  /// [loginWithCredentials].
+  Future<Credentials> confirmPhoneVerificationCode(String phoneNumber, String verificationCode) async {
+    String body = jsonEncode({
+      'authCode': verificationCode,
+      'phoneNumber': phoneNumber,
+    });
+
+    Map<String, dynamic> response = await _httpPost('/api/smartapp/verify/confirm', body);
```
ipainchaud, line 78: It would be nice to have this URL as: `api/smartapp/verify/phone/confirm' or either merge those two into one endpoint. I understand though that this would be a breaking change as I haven'e see any change in platform concerning endpoint URL.

Another one for Later!?!

ipainchaud (approved):

lib/src/models/profile.dart
vsomayaji: True, but Dart seems to encourage grouping things that logically belong together.

https://dart.dev/guides/language/effective-dart/design#consider-declaring-multiple-classes-in-the-same-library

I may reconsider this in the future if it turns into a mess. 🤣

lib/src/platform_client.dart
vsomayaji: Yep, another one for later. I think we'll need an authentication service and will get the opportunity to redesign the APIs.

PR #13, opened 2022-06-11
feat(messaging): support sending and receiving files

ipainchaud:
This PR adds a simple File Sharing capability. It is not complete though as PR15 adds the capability to send message and files prior to a call.

PR #15, opened 2022-06-30
feat: start call with message and file

ipainchaud:
Here I have separated the Messaging from WebRTC's room as they are not related. This makes it much easier to send message before calling.

### DIFF
--- a/lib/src/messaging_client.dart
+++ b/lib/src/messaging_client.dart
@@ -48,7 +48,7 @@ class MessagingClientPubNub implements MessagingClient {
         // from Platform when logging in so: 1) we don't have to provide them to partners; and 2) they can be rotated.
         publishKey: messagingKeys.sendKey,
         subscribeKey: messagingKeys.receiveKey,
-        uuid: pn.UUID(session.userId.toString()),
+        userId: pn.UserId(session.userId.toString()),
       ),
     )
   {
--- a/lib/src/messaging_client.dart
+++ b/lib/src/messaging_client.dart
@@ -18,6 +18,8 @@ abstract class MessagingClient {
   /// If the application does not support messaging, this will throw an exception.
   Stream<Message> get messageStream;
 
+  Future<void> sendStart();
+
   /// Sends the provided message to the Agent.
   ///
   /// If the application does not support messaging, this will throw an exception.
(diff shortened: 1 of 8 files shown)


lib/src/messaging_client.dart
```diff
@@ -0,0 +1,159 @@
...
+abstract class MessagingClient {
+  /// The ID of the service request.
+  set serviceRequestId(int id);
+
+  /// A broadcast stream of messages sent and received.
+  ///
+  /// If the application does not support messaging, this will throw an exception.
+  Stream<Message> get messageStream;
+
+  Future<void> sendStart();
```
ipainchaud, line 21: I forgot to add a comment here... will do! This function, it turns out, is essential for fixing the issue where messages from previous call sneaks into later calls. It is only used by this sdk though and I would like to hide this function from the API. I left this function here for now as I didn't want to delay implementation of the feature.

lib/src/platform_client.dart
```diff
@@ -186,13 +191,52 @@ class PlatformClient {
...
+        List<Future<SentFileInfo>> futureFileInfo = fileMap.entries.map((e) =>
+            messagingClient!.sendFile(e.key, e.value)).toList(growable: false);
+        List<SentFileInfo> fileInfoList = await Future.wait(futureFileInfo);
+
+        List<String> fileIds = fileInfoList.map((fi) => fi.id).toList(growable: false);
+        return fileIds;
+      }
+    } else if (null != text && text.isNotEmpty){
+      await messagingClient!.sendMessage(text);
+    }
```
ipainchaud, line 238: The idea of separating messages this way is to make it easier to manage later into the history. If we have many files, we should be able to delete any of those separately (hence separating them from the message). If we have only one file to send and it is accompanied by a file, I keep them together as the user will most likely want to delete them both if he ever wants to delete them.

lib/src/room.dart
```diff
@@ -60,15 +66,9 @@ abstract class Room implements Listenable {
   /// Sends the provided message to the Agent.
   ///
   /// If the application does not support messaging, this will throw an exception.
+  @Deprecated('This function was moved into [MessagingClient].')
   Future<void> sendMessage(String text);
```
ipainchaud, line 70: I kept this function here as a Deprecated function because this would be a breaking change on the version which were distributed until now.

vsomayaji (approved):
It's a bit difficult to review these changes independently of #13, so I think we should merge this and review the changes together.

PR #20, opened 2022-08-02
feat: add device ID and model to service request context

vsomayaji:
This adds the device ID and model to the service request `context` for requests made from iOS and Android.

On my iPhone Xs, the `model` is the generic string "iPhone", which isn't very helpful. However, at this time, we aren't making any decisions based on the specific iPhone model.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -484,6 +484,8 @@ class PlatformClient {
       };
     } else {
       return {
+        'id': await _deviceId,
+        'model': (await DeviceInfoPlugin().deviceInfo).toMap()['model'],
         'platform': Platform.operatingSystem.toString().split('.').last,
         'platformVersion': Platform.operatingSystemVersion,
       };


ipainchaud (commented):
Would it be helpful to also provide something for WEB. Doesn't need to be something specific to the computer, but the browser could be nice...

vsomayaji:
> Would it be helpful to also provide something for WEB. Doesn't need to be something specific to the computer, but the browser could be nice...

We already get this for web:
```json
{
  "app": {
    "version": "1.0.0+1"
  },
  "device": {
    "platform": "chrome",
    "platformVersion": "5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.5060.134 Safari/537.36"
  },
  "intent": "NONE",
  "permissions": {
    "location": true
  }
}
```

I couldn't think of a value for `id` or `model` that isn't already in there. 🤷🏾

ipainchaud (approved):

ipainchaud:
"facepalm" I had not read the platform param.. good! thanks!

PR #21, opened 2022-08-02
feat: use QoS level 1 for critical messages

vsomayaji:
This switches from MQTT [Quality of Service](https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/) (QoS) level 0 (at most once) to level 1 (at least once) for critical messages that are safe to deliver multiple times.

### DIFF
--- a/lib/src/platform_mq.dart
+++ b/lib/src/platform_mq.dart
@@ -50,7 +50,7 @@ class PlatformMQImpl implements PlatformMQ {
       _client.connectionMessage = MqttConnectMessage()
         ..startClean()
         ..withWillMessage(lastWillMessage)
-        ..withWillQos(MqttQos.atMostOnce)
+        ..withWillQos(MqttQos.atLeastOnce)
         ..withWillTopic(lastWillTopic);
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/platform_mq.dart
```diff
@@ -50,7 +50,7 @@ class PlatformMQImpl implements PlatformMQ {
       _client.connectionMessage = MqttConnectMessage()
         ..startClean()
         ..withWillMessage(lastWillMessage)
-        ..withWillQos(MqttQos.atMostOnce)
+        ..withWillQos(MqttQos.atLeastOnce)
```
vsomayaji, line 53: We want to ensure that the last will message is delivered at least once so a queued service request will be cancelled.

lib/src/room.dart
```diff
@@ -112,7 +112,7 @@ class KurentoRoom extends ChangeNotifier implements Room {
     // Asynchronously subscribe to the room-related topics.
     await _mq.subscribe(_participantEventTopic, MqttQos.atMostOnce, _handleParticipantEventMessage);
     await _mq.subscribe(_participantTopic, MqttQos.atMostOnce, _handleParticipantMessage);
-    await _mq.subscribe(_serviceRequestPresenceTopic, MqttQos.atMostOnce, _handleServiceRequestPresenceMessage);
+    await _mq.subscribe(_serviceRequestPresenceTopic, MqttQos.atLeastOnce, _handleServiceRequestPresenceMessage);
```
vsomayaji, line 115: It's important that we don't miss messages sent to the service request presence topic, since that's how Platform informs us of the service request state changes (`ASSIGNED` => `STARTED` => `END`).

ipainchaud (approved):

PR #22, opened 2022-08-02
feat!: add `RoomHandler.takePhoto`

vsomayaji:
BREAKING CHANGE: Clients must now implement `RoomHandler.takePhoto`. In its simplest form, this can return a frame from the video track like this:

```dart
@override
Future<ByteBuffer> takePhoto() async {
  return _localStream!.getVideoTracks()[0].captureFrame();
}
```

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1,4 +1,5 @@
 import 'dart:convert';
+import 'dart:typed_data';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter_aira/src/models/position.dart';
@@ -21,6 +22,9 @@ import 'sfu_connection.dart';
 abstract class RoomHandler {
   /// Adds the remote stream to an [RTCVideoRenderer].
   Future<void> addRemoteStream(MediaStream stream);
+
+  /// Takes a photo.
+  Future<ByteBuffer> takePhoto();
 }
 
 abstract class Room implements Listenable {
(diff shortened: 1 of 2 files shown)


lib/src/room.dart
```diff
@@ -21,6 +22,9 @@ import 'sfu_connection.dart';
 abstract class RoomHandler {
   /// Adds the remote stream to an [RTCVideoRenderer].
   Future<void> addRemoteStream(MediaStream stream);
+
+  /// Takes a photo.
+  Future<ByteBuffer> takePhoto();
```
ipainchaud, line 27: hehe! I can see why we have to do this, but I like it.

lib/src/room.dart
```diff
@@ -357,7 +350,12 @@ class KurentoRoom extends ChangeNotifier implements Room {
         _log.warning('cannot take photo when presenting');
         return;
       }
-      await _client.uploadPhoto(_serviceRequest.id, await _localStream!.getVideoTracks()[0].captureFrame());
+
+      try {
+        await _client.uploadPhoto(_serviceRequest.id, await _roomHandler.takePhoto());
+      } catch (e, s) {
+        _log.shout('failed to take photo', e, s);
```
ipainchaud, line 357: good thinking on the try catch... thanks!

ipainchaud (approved):

PR #23, opened 2022-08-02
feat: optional video track

vsomayaji:


lib/src/room.dart
```diff
@@ -158,11 +158,11 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
 
-  // The audio is muted if there is no audio track or if the first audio track is disabled.
-  bool get _isAudioMuted => _localStream!.getAudioTracks().isEmpty ? true : !_localStream!.getAudioTracks()[0].enabled;
+  // The audio is muted if the first audio track is disabled.
+  bool get _isAudioMuted => _localStream!.getAudioTracks().isEmpty ? false : !_localStream!.getAudioTracks()[0].enabled;
 
-  // The video is muted if there is no video track or if the first video track is disabled.
-  bool get _isVideoMuted => _localStream!.getVideoTracks().isEmpty ? true : !_localStream!.getVideoTracks()[0].enabled;
+  // The video is muted if the first video track is disabled.
+  bool get _isVideoMuted => _localStream!.getVideoTracks().isEmpty ? false : !_localStream!.getVideoTracks()[0].enabled;
```
vsomayaji, line 165: Since there's no video track, we shouldn't say the video is muted. This is debatable. It depends on whether we want the `videocam_off` icon to show up in Dash.

lib/src/sfu_connection.dart
```diff
@@ -31,17 +33,29 @@ class SfuConnection {
...
+        _video = await _peerConnection.addTransceiver(
+          init: RTCRtpTransceiverInit(direction: TransceiverDirection.SendOnly, streams: [_localStream!]),
+          kind: RTCRtpMediaType.RTCRtpMediaTypeVideo,
+          track: _localStream!.getVideoTracks()[0],
+        );
+      } else {
+        _video = await _peerConnection.addTransceiver(
+          init: RTCRtpTransceiverInit(direction: TransceiverDirection.SendOnly),
+          kind: RTCRtpMediaType.RTCRtpMediaTypeVideo,
+        );
```
vsomayaji, line 58: An empty send-only video transceiver.

vsomayaji (commented):
@ipainchaud did you try something like this? We have to make some changes on the Dash side as to how it determines if video is playing, but (aside from it needing to be done in 3 separate places) that's pretty straightforward.

lib/src/sfu_connection.dart
ipainchaud: I'm actually surprised that the call to addTransceiver doesn't fail, but I like that for its simplicity. Let's go in that direction then.

ipainchaud (approved):

ipainchaud:
I'll test that fully with my changes on the ExplorerUI side and take it from here if you don't mind.

lib/src/sfu_connection.dart
vsomayaji: @ipainchaud the idea was borrowed from https://www.w3.org/TR/webrtc/#advanced-peer-to-peer-example-with-warm-up, which I thought I mentioned to you but perhaps forgot. 🤦🏾 

https://microsoft.github.io/MixedReality-WebRTC/manual/glossary.html also references something similar:

> A transceiver always has exactly one [local media track](https://microsoft.github.io/MixedReality-WebRTC/manual/glossary.html#local-media-track) (sender) and one [remote media track](https://microsoft.github.io/MixedReality-WebRTC/manual/glossary.html#remote-media-track) (receiver), both of which can be null. If null, the transceiver acts as if a dummy track existed which draws its media frames from a null source (black frames for video, silence for audio).

lib/src/room.dart
```diff
@@ -159,10 +160,16 @@ class KurentoRoom extends ChangeNotifier implements Room {
   }
 
   // The audio is muted if the first audio track is disabled.
-  bool get _isAudioMuted => _localStream!.getAudioTracks().isEmpty ? false : !_localStream!.getAudioTracks()[0].enabled;
+  bool get _isAudioMuted => _localStream!.getAudioTracks().isEmpty ? true : !_localStream!.getAudioTracks()[0].enabled;
```
ipainchaud, line 163: I have a fix on dash which I'll commit which makes this experience much better. This let the agent know that the Explorer is not going to share the video in a nice way.

lib/src/room.dart
```diff
@@ -159,10 +160,16 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
 
   // The video is muted if the first video track is disabled.
-  bool get _isVideoMuted => _localStream!.getVideoTracks().isEmpty ? false : !_localStream!.getVideoTracks()[0].enabled;
+  bool get _isVideoMuted {
+    MediaStreamTrack? activeVideoTrack = _activeVideoTrack;
+    return null == activeVideoTrack || !activeVideoTrack.enabled;
+  }
```
ipainchaud, line 169: This allows the manage the mutes in a seamless way. This separates the WebCam and Screenshare mutes from each other.

lib/src/room.dart
```diff
@@ -211,37 +218,48 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
   Future<void> startPresenting(MediaStream displayStream) async {
     // Eventually, we will create a separate connection for screen sharing. That requires changes to Platform and Dash,
     // so for now, we start presenting by replacing the Explorer's video track with the display video track.
-    await _connectionByTrackId[_localTrackId]!.replaceTrack(displayStream.getVideoTracks()[0]);
-    _presenting = true;
-    _log.info('started presenting track=${displayStream.getVideoTracks()[0].label}');
+    _presentationVideoTrack = displayStream.getVideoTracks()[0];
+    await _connectionByTrackId[_localTrackId]!.replaceVideoTrack(_presentationVideoTrack);
+    _log.info('started presenting track=${_presentationVideoTrack!.label}');
+    await _updateParticipantStatus();
```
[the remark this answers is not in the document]
vsomayaji: I think adding mute support for screen sharing is adding unnecessary complexity. Try this in the Aira Desktop app:
1. Start screen sharing.
2. Disable camera. This will mute the screen share and will toggle the button to enable camera. It will also say "camera disabled" even though it's not.
3. Stop sharing screen. Dash will report that the camera is enabled, but it doesn't appear to be.

vsomayaji (commented):
If it was me, I would have created separate PRs for supporting optional video tracks and for supporting muting screen shares, since they're different features and I have reservations about the latter. 🤷🏾

lib/src/room.dart
ipainchaud: this was inspired from a conversation with Troy. If the Explorer wants to handle some personal stuff quickly and would like to hide his screen quickly, stop and start a new screen share is much longer to achieve. Mute would be super quick. This start to be an advanced feature... I thought we could have our Explorer test that out and let us know what they think.

lib/src/room.dart
```diff
@@ -211,37 +218,48 @@ class KurentoRoom extends ChangeNotifier implements Room {
     if (_localStream == null) {
       throw StateError('Cannot mute audio before joining the room');
     }
-
-    _localStream!.getAudioTracks()[0].enabled = !muted;
-    _updateParticipantStatus();
+    if (_localStream!.getAudioTracks().isEmpty) {
+      _log.warning('Cannot mute audio because there is not audio tracks to mute');
```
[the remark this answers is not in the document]
ipainchaud: I'll update.

lib/src/room.dart
```diff
@@ -211,37 +218,48 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
   void setVideoMuted(bool muted) {
     if (_localStream == null) {
       throw StateError('Cannot mute video before joining the room');
     }
-
-    _localStream!.getVideoTracks()[0].enabled = !muted;
-    _updateParticipantStatus();
+    MediaStreamTrack? activeVideoTrack = _activeVideoTrack;
+    if (null == activeVideoTrack) {
+      _log.warning('Cannot mute video because there is not video tracks to mute');
```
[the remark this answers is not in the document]
ipainchaud: I'll update.

lib/src/room.dart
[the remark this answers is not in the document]
ipainchaud: I had in mind to change that message when we are presenting, but forgot. I'll keep my notepad handy to avoid this in the future.

ipainchaud (approved):

PR #27, opened 2022-08-09
fix: `createServiceRequest` fails when messaging is not supported

vsomayaji:


lib/src/platform_client.dart
```diff
@@ -217,7 +217,9 @@ class PlatformClient {
       {Position? position, String? message, Map<String, List<int>>? fileMap, bool? cannotTalk}) async {
     _verifyIsLoggedIn();
 
-    List<String> fileIds = await _sendPreCallMessage(message, fileMap);
+    if ((message?.isNotEmpty ?? false) || (fileMap?.isNotEmpty ?? false)) {
+      await _sendPreCallMessage(message, fileMap);
```
vsomayaji, line 221: `_sendPreCallMessage` throws an exception if the application does not support messaging.

lib/src/platform_client.dart
```diff
@@ -233,7 +235,6 @@ class PlatformClient {
       'requestType': 'AIRA', // Required but unused.
       'hasMessage': (null != message && message.isNotEmpty) || fileNames.isNotEmpty || true == cannotTalk,
       'message': '$message${fileNames.isEmpty ? '' : ' (With files: $fileNames)'}',
-      'fileIds': fileIds,
```
vsomayaji, line 236: I'm not sure why we were sending these to Platform, as it doesn't do anything with them.

ipainchaud: good catch! At first, I wanted to make Dash retrieve the URLs of the sent files within the dialog shown at call start. After having seen that this change in Dash wasn't as trivial, I decided to leave the filenames only on that dialog but forgot to remove the fileIds.

ipainchaud (approved):

PR #35, opened 2022-08-15
feat!: simplified mutes

### LINKED ISSUE #34: Simplified mutes
There are a few things about our mutes implementation that don't sit well with me:

1. There's a different way to mute a track _before_ joining a `Room`:
   ```dart
   _localMediaStream.getAudioTracks()[0].enabled = false;
   ```
   And _after_ joining a `Room`:
   ```dart
   _room.setAudioMuted(true);
   ```
   If you use the first way _after_ joining a `Room`, the mute states of the track and the `Room` will be out of sync.

2. The `Room` does not expose getters for its mute states. If the mute states of the track and the `Room` get out of sync, there's no way to know.

3. `Room.setAudioMuted` and `Room.setVideoMuted` perform an async call (to `_updateParticipantStatus`) but do not return a `Future`.

4. `Room.setVideoMuted` only mutes the _active_ video track. Replacing the video track (with `Room.replaceStream`) will not preserve the mute state.

5. `Room.setVideoMuted` affects the presentation even though the presentation was intended to be thought of as a separate video track.

Now that we've discovered that we can "mute" a track by providing the corresponding transceiver a `null` track, I think we can switch to an implementation where the mute states are always managed by the `Room`. The interface could look like this:

```dart
  /// Whether the local audio is muted.
  bool get isAudioMuted;

  /// Mutes or un-mutes the local audio.
  Future<void> setAudioMuted(bool muted);

  /// Whether the local video is muted.
  bool get isVideoMuted;

  /// Mutes or un-mutes the local video.
  Future<void> setVideoMuted(bool muted);

  /// Whether the presentation is muted.
  bool get isPresentationMuted;

  /// Mutes or un-mutes the presentation.
  Future<void> setPresentationMuted(bool muted);

  /// Enables or disables privacy mode.
  ///
  /// In privacy mode, the local audio, video and presentation are _all_ muted. This is equivalent to calling
  /// [setAudioMuted], [setVideoMuted] and [setPresentationMuted], and is intended as a convenience for apps that do not
  /// support muting the tracks independently.
  Future<void> setPrivacyMode(bool enabled);
```

I believe this would address all of the issues above. For example, if the video is muted in the `Room` and the video track is replaced, the video will stay muted.

What do you think? Good idea or a bad one?

vsomayaji:
Resolves #34.

BREAKING CHANGE: The `Room` now manages its mute states separately from the `MediaStreamTrack`s. Previously, to join a `Room` with the audio muted, you would do this before calling `Room.join`:
```dart
_localMediaStream.getAudioTracks()[0].enabled = false;
``` 

Now, you will do this:
```dart
_room.setAudioMuted(true);
```

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -51,9 +51,6 @@ abstract class Room implements Listenable {
   Stream<Message> get messageStream;
 
   /// Joins the room with the provided local audio and video stream.
-  ///
-  /// If the room should be joined with the audio and/or video muted, disable the corresponding track(s) before calling
-  /// [join].
   Future<void> join(MediaStream localStream);
 
   /// Whether the local audio is muted.
@@ -68,17 +65,21 @@ abstract class Room implements Listenable {
   /// Mutes or un-mutes the local video.
   Future<void> setVideoMuted(bool muted);
 
+  /// Whether the presentation is muted.
+  bool get isPresentationMuted;
+
+  /// Mutes or un-mutes the presentation.
+  Future<void> setPresentationMuted(bool muted);
+
   /// Enables or disables privacy mode.
   ///
-  /// In privacy mode, both the local audio and video (if present) are muted. This is equivalent to calling both
-  /// [setAudioMuted] and [setVideoMuted] and is intended as a convenience for apps that do not support muting the audio
-  /// and video independently.
+  /// In privacy mode, the local audio, video and presentation are _all_ muted. This is equivalent to calling
+  /// [setAudioMuted], [setVideoMuted] and [setPresentationMuted], and is intended as a convenience for apps that do not
+  /// support muting the tracks independently.
   Future<void> setPrivacyMode(bool enabled);
 
   /// Starts presenting the provided display stream.
   Future<void> startPresenting(MediaStream displayStream);
-
-  Future<void> setPresentationMuted(bool muted);
 
   /// Stops presenting the display stream.
   Future<void> stopPresenting();
@@ -118,7 +119,9 @@ class KurentoRoom extends ChangeNotifier implements Room {
   bool _isDisposed = false;
   bool _isAudioMuted = false;
   bool _isVideoMuted = false;
+  bool _isPresentationMuted = false;
   MediaStreamTrack? _presentationVideoTrack;
+
   bool get _isPresenting => null != _presentationVideoTrack;
   ServiceRequestState _serviceRequestState = ServiceRequestState.queued;
   String? _agentName;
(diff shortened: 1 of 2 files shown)


lib/src/room.dart
```diff
@@ -51,16 +51,32 @@ abstract class Room implements Listenable {
...
   /// Joins the room with the provided local audio and video stream.
-  ///
-  /// If the room should be joined with the audio and/or video muted, disable the corresponding track(s) before calling
-  /// [join].
   Future<void> join(MediaStream localStream);
 
-  /// Mutes or un-mutes the local audio stream.
-  void setAudioMuted(bool muted);
+  /// Whether the local audio is muted.
+  bool get isAudioMuted;
```
vsomayaji, line 57: Since the `Room` now manages its own mute states, we expose getters to retrieve the mute states: `isAudioMuted`, `isVideoMuted` and `isPresentationMuted`.

lib/src/room.dart
```diff
@@ -51,16 +51,32 @@ abstract class Room implements Listenable {
...
+
+  /// Mutes or un-mutes the local audio.
+  Future<void> setAudioMuted(bool muted);
```
vsomayaji, line 60: These now return a `Future` so they can be `await`ed.

lib/src/room.dart
```diff
@@ -51,16 +51,32 @@ abstract class Room implements Listenable {
...
+
+  /// Mutes or un-mutes the presentation.
+  Future<void> setPresentationMuted(bool muted);
+
+  /// Enables or disables privacy mode.
+  ///
+  /// In privacy mode, the local audio, video and presentation are _all_ muted. This is equivalent to calling
+  /// [setAudioMuted], [setVideoMuted] and [setPresentationMuted], and is intended as a convenience for apps that do not
+  /// support muting the tracks independently.
+  Future<void> setPrivacyMode(bool enabled);
```
vsomayaji, line 79: This is for mobile or glasses apps that have a single "privacy mode" toggle. This allows us to call `_updateParticipantStatus` once instead of once per mute.

lib/src/sfu_connection.dart
```diff
@@ -14,15 +24,20 @@ class SfuConnection {
...
-      [MediaStream? localStream])
-      : _localStream = localStream;
+  SfuConnection({
+    required this.direction,
+    required this.onConnectionState,
+    required this.onIceCandidate,
+    required this.onSdpOffer,
+    required this.onTrack,
+    required this.trackId,
+  });
```
ipainchaud, line 34: it would have been nice to take this occasion to send the constructor as the first defined element of the class.

ipainchaud (approved):

PR #39, opened 2022-09-01
feat: `getUser`

BenReber:
This adds the API `PlatformClient.getUser`. It also modifies `PlatformClient.createServiceRequest` to accept an optional `accountId`.

The example has been modified to demonstrate how these changes can be used to implement profile switching.

Screen Shot 2022-09-12 at 3 56 52 PM


lib/src/models/service_request.dart
```diff
@@ -17,6 +17,7 @@ class ServiceRequest {
   final int participantId;
   final int roomId;
   final int userId;
+  final int accountId;
```
vsomayaji, line 20: We don't need to modify this class.

lib/src/platform_client.dart
```diff
@@ -585,3 +585,29 @@ class PlatformMessagingKeys {
 
   PlatformMessagingKeys(this.sendKey, this.receiveKey);
 }
+
+/// Get User information from API and return relevant information on the account.
+Future<Session> getUserDetails(String token, int userId) async {
```
vsomayaji, line 590: We don't need to pass in either the `token` or `userId`. The `PlatformClient` has both. We also need to return a different model (`Session` doesn't have what we need).

vsomayaji (commented):
We need to modify `createServiceRequest` to accept an optional `accountId`.

lib/src/platform_client.dart
```diff
@@ -216,7 +217,7 @@ class PlatformClient {
   ///
   /// [cannotTalk] will let the agent know if the Explorer cannot talk at connection time.
   Future<Room> createServiceRequest(RoomHandler roomHandler,
-      {Position? position, String? message, Map<String, List<int>>? fileMap, bool? cannotTalk}) async {
+      {Position? position, String? message, Map<String, List<int>>? fileMap, bool? cannotTalk, int? accountId}) async {
```
BenReber, line 220: Add optional account ID so billing can be charged based on account ID with profile

vsomayaji (approved):

PR #40, opened 2022-09-08
update: updated to flutter 3.3

ipainchaud:
updated for flutter 3.3 and resolving issue with flutter_webrtc which uses the latest webrtc_interface which is no longer compatible

PR #41, opened 2022-09-08
feat: reconnect

vsomayaji:
This adds support for handling `RECONNECT` messages from the Agent. On a reconnect, we close all of our connections with the SFU and create new ones.

To get notified when a reconnect is initiated and completed, an app can use the new `Room.onReconnect` and `Room.onReconnected` callbacks.


example/lib/main.dart
```diff
@@ -419,34 +419,30 @@ class _MyAppState extends State<MyApp> implements RoomHandler {
           // Rebuild the UI.
           setState(() {});
 
-          ScaffoldMessenger.of(context).showSnackBar(SnackBar(
-            content: Text('Connected to Agent ${_room!.agentName}'),
-          ));
+          _showSnackBar('Connected to Agent ${_room!.agentName}');
```
vsomayaji, line 422: Saved a couple lines by creating a helper.

example/pubspec.lock
```diff
@@ -180,7 +180,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.0.21"
+    version: "0.0.22"
```
vsomayaji, line 183: This was missed in #40.

ipainchaud: dang! I'm taking note to run `pub get` before committing next time.

lib/src/room.dart
```diff
@@ -581,4 +578,60 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+      _connectionByTrackId.clear();
+
+      // Create and connect new tracks to receive the remote streams from the other participant(s). (Do this before our
+      // local stream, since we want to be able to hear the Agent ASAP.)
+      for (Participant participant in await _client.getParticipants(_serviceRequest.roomId)) {
+        if (participant.id != _serviceRequest.participantId) {
+          for (Track track in participant.tracks ?? []) {
+            if (track.isOutgoing) {
+              await _createIncomingTrack(track.id);
+            }
```
ipainchaud, line 622: I can see how this could make sense... it would be nice if all the functions were using the same point of reference. Here, the if condition `isOutgoing` seems to contradict the action `createIncomingTrack`.

Maybe the best solution to this issue is to use an "absolute" nomenclature (opposed to "relative" nomenclature) like `trackFromAgent` and `trackToAgent`.

vsomayaji: This is a side effect of how things are modeled in Platform. Tracks are either incoming or outgoing. An outgoing track is a send-only track from the participant to Kurento. An incoming track is a receive-only track from Kurento to the participant that maps to an outgoing track from another participant.

PR #43, opened 2022-09-09
build: release 0.0.23

vsomayaji:


example/ios/Podfile
```diff
@@ -37,5 +37,10 @@ end
 post_install do |installer|
   installer.pods_project.targets.each do |target|
     flutter_additional_ios_build_settings(target)
+
+    target.build_configurations.each do |config|
+      # https://stackoverflow.com/a/70632707
+      config.build_settings.delete 'IPHONEOS_DEPLOYMENT_TARGET'
+    end
```
vsomayaji, line 44: `flutter build ipa` was failing with:
```
fatal error: 'WebRTC/WebRTC.h' file not found
```
This was the only solution I found that seemed to resolve it.

ipainchaud: oddly enough! it doesn't fail on xcode when I compile there...

ipainchaud (approved):

PR #44, opened 2022-09-12
feat: throttle `Room.updateLocation`

ipainchaud:


lib/src/room.dart
```diff
@@ -289,6 +290,11 @@ class KurentoRoom extends ChangeNotifier implements Room {
       return;
     }
 
+    const int delta = 2 * 1000; // 2 seconds.
+    if (DateTime.now().millisecondsSinceEpoch - _lastPositionUpdate < delta) {
+      return;
+    }
```
vsomayaji, line 296: A few things:
1. How did you settle on 2 seconds? Let's start with 1 and see how that goes. (If we have a spammy app, that should be investigated. If you're not moving your phone, that shouldn't trigger a location update. Otherwise, we're using compute -- and draining the battery -- when we don't need to.)
2. `difference` seems like a cleaner way to do date math.
3. To keep location updates from stacking up, we should set the updated date as soon as possible instead of after the update.

How about:

```dart
DateTime _lastLocationUpdate = DateTime.fromMillisecondsSinceEpoch(0);
```

```dart
    DateTime now = DateTime.now();
    if (now.difference(_lastLocationUpdate).inSeconds < 1) {
      return;
    }
    _lastLocationUpdate = now;
```

vsomayaji (approved):

PR #49, opened 2022-10-11
feat: usage

ipainchaud:


lib/src/models/profile.dart
```diff
@@ -31,7 +31,15 @@ class Profile {
...
+  final int durationAllowed;
+  /// Duration Used for this Account in minutes
+  final int durationUsed;
+
   Profile.fromJson(Map<String, dynamic> json)
       : account = Account.fromJson(json['account']),
-        type = ProfileType.fromName(json['userType']);
+        type = ProfileType.fromName(json['userType']),
+        durationAllowed = _convertSecondsToMinutes(json['durationAllowed']),
+        durationUsed = _convertSecondsToMinutes(json['durationUsed']);
```
vsomayaji, line 43: This doesn't feel like the right place to expose this data. A couple other concerns:
1. This means you'll have to refresh the profile after every service request, since the `durationUsed` could have changed.
2. Converting to minutes means you're losing information. Platform will allow you to make a call if you have at least 10 seconds left, but that'll show up as 0 minutes.

lib/src/models/user.dart
```diff
@@ -5,10 +5,12 @@ class User {
   final String firstName;
   final String lastName;
   final List<Profile> profiles;
+  final bool isGuest;
 
   User.fromJson(Map<String, dynamic> json)
       : id = json['id'],
         firstName = json['firstName'],
         lastName = json['lastName'] ?? '',
+        isGuest = json['guest'] ?? true, // This information comes from table `user_property_value` and defaults to true in platform (see SubscriptionManager.isGuestSafe(userId))
```
vsomayaji, line 14: I'm not a fan of this and hope we can eliminate it in a future refactor. If we're trying to determine whether a user has a plan/subscription, we can use the usage data for that.

vsomayaji (approved):

vsomayaji (approved):

lib/src/models/user.dart
ipainchaud: I didn't know that this information could be found in usage... Is this the `lastName` that you are referring to? If this is the case, this is not super pretty. Also, why don't you like this information with the User information. This, to me, seems to go well together.

lib/src/models/profile.dart
ipainchaud: I converted to minutes to keep consistency with the other API which return the time for the user profile (in minutes). As for the place... I agree. This profile information is a little "all over the place" to my eyes and I didn't think this was the time to mingle with this.

Thanks for your comment though as I realize we have an issue where the Usage should read the profile information from the API before displaying to ensure fresh data.

PR #50, opened 2022-10-24
feat: add `PlatformClient.registerPushToken`

vsomayaji:


lib/src/platform_client.dart
```diff
@@ -433,6 +433,26 @@ class PlatformClient {
...
+  Future<void> registerPushToken(String token) async {
+    _verifyIsLoggedIn();
+
+    if (kIsWeb || !(Platform.isAndroid || Platform.isIOS)) {
+      throw UnsupportedError('Unsupported platform');
+    }
+
+    String body = jsonEncode({
+      'token': token,
+      'type': Platform.isAndroid ? 'ANDROID' : 'IOS',
```
ipainchaud, line 450: would be nice to check if the platform is supported... I don't think that expecting that all users of this API will check this appropriately is super wise.

ipainchaud (approved):

lib/src/platform_client.dart
vsomayaji: I don't understand this comment. Above, we throw an `UnsupportedError` if the platform is not supported:
```dart
    if (kIsWeb || !(Platform.isAndroid || Platform.isIOS)) {
      throw UnsupportedError('Unsupported platform');
    }
```

Should we be doing more?

PR #52, opened 2022-10-31
feat: add `User.email` and `User.phoneNumber`

ipainchaud:


lib/src/models/usage.dart
```diff
@@ -41,27 +41,34 @@ class PlanUsageBreakdown {
 
 /// Contains main information about usage.
 class Usage {
+
```
ipainchaud, line 44: Sorted the properties from that Model. No other change.

lib/src/models/user.dart
```diff
@@ -2,15 +2,17 @@ import 'profile.dart';
 
 class User {
   final int id;
+  final String? email;
   final String firstName;
   final String lastName;
+  final String? phoneNumber;
   final List<Profile> profiles;
-  final bool isGuest;
```
ipainchaud, line 8: Removed unused `isGuest` and added `email` and `phoneNumber` because they are used by the SSO when we upgrade a plan.

lib/src/models/usage.dart
```diff
@@ -41,27 +41,34 @@ class PlanUsageBreakdown {
...
+  /// Indicator if the Explorer has access to a 5 min free call.
+  bool? hasFreeCall;
+
   /// Name of the plan.
   String? planName;
 
-  /// Total available minutes in the plan.
-  int? totalMinutes;
+  /// Indicates if the plan is unlimited.
+  bool? planUnlimited;
```
vsomayaji, line 58: I think we will be getting rid of unlimited plans. If we don't, maybe we can eventually eliminate this and instead set `totalPrimaryMinutes` to `-1` (a pattern for "unlimited" used elsewhere in Platform).

lib/src/models/usage.dart
```diff
@@ -41,27 +41,34 @@ class PlanUsageBreakdown {
...
 
-  /// Total of used minutes in the plan.
-  int? totalMinutesLeft;
+  /// Indicates if the User is Primary.
+  bool primary;
```
vsomayaji, line 61: One day, we could change this to `isPrimary`. Or maybe replace the `bool` with a `ProfileType`, which already defines `primary` and `secondary`.

lib/src/models/usage.dart
```diff
@@ -41,27 +41,34 @@ class PlanUsageBreakdown {
...
 
-  /// Total amount of primary minutes in the plan. These primary minutes are personal minutes.
-  int? totalPrimaryMinutes;
+  /// Breakdown of the usage by Primary and Secondary users.
+  List<PlanUsageBreakdown>? planUsageBreakdownList;
```
vsomayaji, line 64: `PlanUsageBreakdown` is an odd name. Same for `planUsageBreakdownList`.

lib/src/models/usage.dart
```diff
@@ -41,27 +41,34 @@ class PlanUsageBreakdown {
...
 
-  /// Total amount of used primary minutes in the plan.
-  int? totalPrimaryMinutesUsed;
+  /// Delay before next free 5 min call.
+  DateTime? timeUntilNextFreeCall;
```
vsomayaji, line 67: `timeUntil` suggests that this is a delta but I think it's the actual time that the next free call becomes available. Maybe `nextFreeCallDate` would be more appropriate. 

Also, I noticed that we sometimes name `DateTime` properties as "TimeStamp" (e.g. `CallSession.endTimeStamp`). It'd be nice if we used a consistent naming convention. My lean is toward "Date"(`Usage.billingCycleEndDate`, `Usage.nextFreeCallDate`, `CallSession.endDate`) as I was using "At" for dates passed as milliseconds since the epoch (`Message.sentAt`).

lib/src/models/usage.dart
```diff
@@ -74,26 +81,20 @@ class Usage {
...
 
-  /// Delay before next free 5 min call.
-  DateTime? timeUntilNextFreeCall;
+  /// Total of used minutes in the plan.
+  int? totalMinutesLeft;
 
-  /// Indicates if the User is Primary.
-  bool primary;
+  /// Grand total of minutes used.
+  int? totalMinutesUsed;
```
vsomayaji, line 91: Maybe we should eliminate the grand totals (`totalMinutes`, `totalMinutesLeft`, `totalMinutesUsed`) as they can be calculated from the primary, credit, and fixed values.

vsomayaji (approved):
I added some comments about the `Usage` class since I hadn't reviewed that before. All stuff for future refactors.

PR #54, opened 2022-11-09
feat(profile): added api for profile data updates.

ipainchaud:


lib/src/models/user.dart
```diff
@@ -4,15 +4,47 @@ class User {
...
+  }) =>
+      User(
+        id: id ?? this.id,
+        email: email ?? this.email,
+        firstName: firstName ?? this.firstName,
+        languages: languages ?? this.languages,
+        lastName: lastName ?? this.lastName,
+        phoneNumber: phoneNumber ?? this.phoneNumber,
+        profiles: profiles ?? this.profiles,
+      );
```
ipainchaud, line 49: I was not sure you'd like that `cloneWith`... I kind of prefer that to making the class mutable... will give an incentive to the users to make sure they really want mutate... plus, we get all the immutability perks.

lib/src/platform_client.dart
```diff
@@ -368,35 +367,31 @@ class PlatformClient {
   Future<void> uploadPhoto(int serviceRequestId, ByteBuffer photo) async {
```
ipainchaud, line 367: that uploadPhoto function wasn't following the usual pattern in error management.

vsomayaji (approved):

PR #60, opened 2022-12-30
feat(AccessOffer): added renewalTimestamp

ipainchaud:
This is required to pull the renewal time of the offer.


vsomayaji:
@ipainchaud we should not be using the `renewalTimestamp`. AFAIK that's a deprecated feature.

PR #61, opened 2023-01-19
fix!: no `start` message sent when call is not started with a message or file

ipainchaud:
BREAKING CHANGE: `MessagingClient.sendStart` has been removed and `PlatformClient.messagingClient` is now read-only.

### DIFF
--- a/lib/src/messaging_client.dart
+++ b/lib/src/messaging_client.dart
@@ -36,6 +36,7 @@ abstract class MessagingClient {
 
 class MessagingClientPubNub implements MessagingClient {
   MessagingClientPubNub(PlatformMessagingKeys messagingKeys, int userId, String token) : _userId = userId {
+    _log.finest('Initializing PubNub...');
     _pubnub = pn.PubNub(
       defaultKeyset: pn.Keyset(
         // TODO: Setting the `authKey` shouldn't be necessary, but there is a bug in the PubNub SDK where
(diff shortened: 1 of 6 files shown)


lib/src/messaging_client.dart
```diff
@@ -16,8 +18,6 @@ abstract class MessagingClient {
   /// If the application does not support messaging, this will throw an exception.
   Stream<Message> get messageStream;
 
-  Future<void> sendStart();
```
vsomayaji, line 19: Removed `sendStart` from the public API, as it's a legacy solution that we want to eliminate.

lib/src/platform_client.dart
```diff
@@ -46,6 +46,7 @@ class PlatformClient {
   /// After this is called, the object is not in a usable state and should be discarded.
   void dispose() {
     _httpClient.close();
+    messagingClient?.dispose();
```
vsomayaji: This was missed in a prior PR.

lib/src/platform_client.dart
```diff
@@ -29,12 +29,14 @@ class PlatformClient {
 
   final PlatformClientConfig _config;
   Session? _session;
-  MessagingClient? messagingClient;
```
vsomayaji, line 32: This should never have been writable. In fact, I don't think it should be accessible from `PlatformClient`, since, from an app's perspective, messaging is only possible in a `Room`. Something for a future refactor...

lib/src/platform_client.dart
```diff
@@ -46,6 +48,7 @@ class PlatformClient {
   /// After this is called, the object is not in a usable state and should be discarded.
   void dispose() {
     _httpClient.close();
+    _messagingClient?.dispose();
```
vsomayaji, line 51: This was missed in a previous PR.

lib/src/room.dart
```diff
@@ -124,7 +123,6 @@ class KurentoRoom extends ChangeNotifier implements Room {
   String? _agentName;
   MediaStream? _localStream;
   int? _localTrackId;
-  pn.Subscription? _messageSubscription;
```
vsomayaji, line 127: This was missed in a previous PR.

vsomayaji (approved):

PR #67, opened 2023-02-07
feat: added rideshare api

ipainchaud:
3 new functions to handle Lyft account linking: getLyftAuthorizationUrl, sendLyftAuthorizationCode and revokeLyftAuthorization

### DIFF
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -172,7 +172,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.0.37"
+    version: "0.0.38"
   flutter_lints:
     dependency: "direct dev"
     description:
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -172,7 +172,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.0.36"
+    version: "0.0.37"
   flutter_lints:
     dependency: "direct dev"
     description:
@@ -549,5 +549,5 @@ packages:
     source: hosted
     version: "6.2.2"
 sdks:
-  dart: ">=2.18.0 <4.0.0"
+  dart: ">=2.18.0 <3.0.0"
   flutter: ">=3.0.0"
--- a/lib/flutter_aira.dart
+++ b/lib/flutter_aira.dart
@@ -14,7 +14,7 @@ export 'src/models/minute_sharing.dart' show MinuteSharingMember, MinuteSharingI
 export 'src/models/service_request.dart' show ServiceRequest, ServiceRequestState;
 export 'src/models/session.dart' show Session;
 export 'src/models/track.dart' show Track;
-export 'src/models/user.dart' show User;
+export 'src/models/user.dart' show User, LinkedAccount;
 export 'src/models/usage.dart' show Usage, PlanUsageBreakdown;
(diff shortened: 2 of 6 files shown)


example/pubspec.lock
```diff
@@ -172,7 +172,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "0.0.36"
+    version: "0.0.37"
```
ipainchaud, line 175: automatically upgraded through "flutter pub upgrade"

lib/flutter_aira.dart
```diff
@@ -14,7 +14,7 @@ export 'src/models/minute_sharing.dart' show MinuteSharingMember, MinuteSharingI
 export 'src/models/service_request.dart' show ServiceRequest, ServiceRequestState;
 export 'src/models/session.dart' show Session;
 export 'src/models/track.dart' show Track;
-export 'src/models/user.dart' show User;
+export 'src/models/user.dart' show User, LinkedAccount;
```
ipainchaud, line 17: We now need to have more information to know which account is linked. We currently only handle Lyft, but in the past there was Uber as well... We may need to refactor this, but the backend work that way currently and I thought it was ok to leave it as is for the moment.

lib/src/models/user.dart
```diff
@@ -23,8 +25,9 @@ class User {
       : id = json['id'],
         email = json['email'],
         firstName = json['firstName'] ?? '',
-        languages = (json['language'] ?? [] as List<String>).map((v) => Language.fromString(v)).cast<Language>().toList(),
+        languages = (json['language'] ?? [] as List<String>).map((v) => Language.fromString(v)).cast<Language>().toList(growable: false),
```
ipainchaud, line 28: I believe this was left as "growable: true" to make it easier to add or remove languages, but I added since the `cloneWith` function which makes it a lot easier to work with immutability.

I'll mark this as a breaking change in the new API version.

lib/src/models/user.dart
```diff
@@ -44,7 +48,36 @@ class User {
...
+      : name = json['serviceName'],
+        accessToken = json['accessToken'],
+        refreshToken = json['refreshToken'],
+        externalId = json['externalId'];
+
+  LinkedAccount cloneWith({
+    String? name,
+    String? accessToken,
+    String? refreshToken,
+    String? externalId,
```
vsomayaji, line 75: I'm not sure we should expose these values to the app. Do you know if/how they're used?

lib/src/platform_client.dart
```diff
@@ -731,6 +731,66 @@ class PlatformClient {
...
+      headers: headers,
+    );
+    if (response.statusCode != 200) {
+      Map<String, dynamic> json = jsonDecode(response.body);
+      throw PlatformLocalizedException(json['response']?['errorCode'], json['response']['errorMessage']);
+    }
+  }
+
+  /// Unregisters LYFT from the user's account.
+  Future<void> revokeLyftAuthorization() {
```
vsomayaji, line 782: Should this be an `async` function?

vsomayaji (dismissed):

lib/src/models/user.dart
ipainchaud: I knew you would ask 😄 ! I should have left a comment there. It is not used AT THE MOMENT. I'll remove them.

lib/src/platform_client.dart
ipainchaud: nop! we don't await, we just return a Future.

PR #68, opened 2023-02-10
feat: debounce reconnects

vsomayaji:
While triaging an [error](https://aira-io.sentry.io/share/issue/34fe6fb73a204a1b9507e020201cad69/) in Sentry, I found that two reconnects were in progress simultaneously. Since that's never desired, we now try and limit reconnects to one at a time.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -116,6 +116,7 @@ class KurentoRoom extends ChangeNotifier implements Room {
   bool _isAudioMuted = false;
   bool _isVideoMuted = false;
   bool _isPresentationMuted = false;
+  bool _isReconnecting = false;
   MediaStreamTrack? _presentationVideoTrack;
 
   bool get _isPresenting => null != _presentationVideoTrack;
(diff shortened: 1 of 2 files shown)


lib/src/room.dart
```diff
@@ -606,6 +607,12 @@ class KurentoRoom extends ChangeNotifier implements Room {
   }
 
   Future<void> _reconnect() async {
+    if (_isReconnecting) {
+      _log.warning('a reconnect is already in progress');
+      return;
+    }
+    _isReconnecting = true;
+
```
ipainchaud, line 615: Does Dashboard have to do anything in case of reconnections? I assume not, but if it does, this could cause an out of synch state. I'll double check... unless you already know! It would be nice to prevent this on the Dashboard side too.

ipainchaud (approved):

lib/src/room.dart
vsomayaji: @ipainchaud Dash just sends a `RECONNECT` event, similar to what it does when taking a photo. It doesn't wait for a response. 

Dash already has a reconnect throttle built in, but in this particular case, it looks like it either didn't work or a message got delivered twice

PR #70, opened 2023-02-13
fix: missed service request status messages

vsomayaji:
We're still missing service request status messages, resulting in errors like [this one](https://aira-io.sentry.io/share/issue/38587b9586e24fe7baeeef023146c6c1/) in Sentry:

> missed service request status message topic=prod/user/1170/service-request/presence

We now periodically call a Platform API to get the service request status and use the value if we miss a message.


lib/src/platform_client.dart
```diff
@@ -305,6 +306,24 @@ class PlatformClient {
...
+  Future<Map<String, String?>> getServiceRequestStatus(int serviceRequestId) async {
+    _verifyIsLoggedIn();
+
+    Map<String, dynamic> response =
+        await _httpGet('/api/user/service', queryParameters: {'id': serviceRequestId.toString()});
+
+    return {
+      'agentFirstName': response['agentName']?.toString().split(' ').first, // Split the first name and last initial.
+      'status': response['serviceStatus'],
+    };
```
vsomayaji, line 324: At the moment, we just return the `agentFirstName` and `status`. In the future, we could potentially return things like the position in queue and the expected wait time. 🤔

lib/src/room.dart
```diff
@@ -141,6 +143,10 @@ class KurentoRoom extends ChangeNotifier implements Room {
     await _mq.subscribe(_participantEventTopic, MqttQos.atMostOnce, _handleParticipantEventMessage);
     await _mq.subscribe(_participantTopic, MqttQos.atMostOnce, _handleParticipantMessage);
     await _mq.subscribe(_serviceRequestPresenceTopic, MqttQos.atLeastOnce, _handleServiceRequestPresenceMessage);
+
+    // HACK: The `_serviceRequestPresenceTopic` has been unreliable and we haven't yet figured out why. Until then,
+    // we're backing it up by periodically checking the status of the service request.
+    _getServiceRequestStatusTimer = Timer.periodic(const Duration(seconds: 3), (_) => _getServiceRequestStatus());
```
vsomayaji, line 149: 3 seconds seems like a reasonable not-too-frequent but not-too-infrequent value, but I'm open to suggestions.

lib/src/room.dart
```diff
@@ -396,17 +404,6 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
-        if (serviceRequestState == ServiceRequestState.queued) {
-          // HACK: If the Agent is sending audio and we still think we're queued, we're not receiving messages on the
-          // service request presence topic. Until we can figure out why this is happening, pretend we received a
-          // message and transition the service request status to started.
-          _log.shout('missed service request status message topic=$_serviceRequestPresenceTopic');
-          _agentName = '';
-          _serviceRequestState = ServiceRequestState.started;
-          await _updateParticipantStatus();
-          notifyListeners();
-        }
```
vsomayaji, line 409: This was buggy for a couple of reasons:
1. We would transition from `queued` to `started` and skip `assigned`, so if an app has any logic that triggers on `assigned` (which the Aira Explorer app currently does), it wouldn't execute.
2. We didn't have the Agent name, so if an app has any logic that triggers on the presence of the Agent name (which the Aira Explorer app currently does), it wouldn't execute.

lib/src/room.dart
```diff
@@ -425,36 +422,61 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+    switch (status) {
+      case 'REQUEST':
+        // Nothing to do.
+        break;
 
-    switch (json['status']) {
       case 'ASSIGNED':
-        _serviceRequestState = ServiceRequestState.assigned;
-        _agentName = json['agentFirstName'];
+        if (_serviceRequestState == ServiceRequestState.queued) {
```
vsomayaji, line 447: We now handle invalid state transitions. For example, it's not valid to go from `started` to `assigned`.

lib/src/room.dart
```diff
@@ -425,36 +422,61 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+        }
         break;
 
       case 'STARTED':
-        _serviceRequestState = ServiceRequestState.started;
+        if (_serviceRequestState == ServiceRequestState.queued) {
+          // HACK: If we still think we're queued, we missed a message. Notify listeners that we're assigned, and let
+          // our periodic timer change this to started the next time it fires.
+          _log.warning('missed service request status message topic=$_serviceRequestPresenceTopic');
+          await _updateServiceRequestStatus('ASSIGNED', agentFirstName);
```
vsomayaji, line 459: This means we'll be a little delayed in getting to `started`, but we will have transitioned states correctly.

ipainchaud: This logic is complicated to understand. I'd need more time to figure out if there is no bug here, but you are usually pretty thorough, so instead of making the family wait me for dinner any longer, I thought I'd ask you the following questions:

I was wondering why, if we missed the 'ASSIGNED` state, don't we simply not call `_getServiceRequestStatus` here? We could wait for the result and then set the state to `STARTED`. This would be much simpler than relying on a timer.

My understanding on how the messages work is quite limited here. Waiting to call here might mean we won't be able to get the agent's name anymore (you say the name comes with the "ASSIGNE" message on a comment above). Not only calling every 3 seconds might miss that window as well, but that whole condition could be useless??? I feel a simple and efficient solution could ensure that the agent name is sent on 'ASSIGNED' and 'STARTED' event?

vsomayaji: @ipainchaud let's huddle in the morning, as I think it'll be easier to thoroughly explain how this works in person.

In short, if we're not receiving messages on the service request presence topic, we're not going to receive _any_ messages on that topic. In the early days of the Aira Explorer app, this is what would cause service requests to be stuck "Waiting for Agent" forever, even while the Agent was actually connected. We had "fixed" that with a hack that forced the state to `started` if we received the Agent's audio and were still `queued`, but that introduced another bug: because the Aira Explorer app now delays joining the room until the state is `assigned`, the Explorer never joins the room if messages are missed. 🤦🏾 Anyway, I'm sure you have some great ideas and I'll discuss them with you tomorrow.

vsomayaji: A few more notes to set up our discussion tomorrow:

* The service request presence topic is unreliable and we haven't yet been able to figure out why. One theory is that this is due to Platform publishing messages with [QoS](https://www.hivemq.com/blog/mqtt-essentials-part-6-mqtt-quality-of-service-levels/) 0 (i.e. at most once), so missed messages are expected. If so, this would affect all topics; we just notice it more for this one because it controls the very visible service request state transitions.
* The service request presence topic is only used for state transitions. We could eliminate it and use the periodic timer instead; however, I opted to keep the topic and use the periodic timer as a backup because the topic requires less resources and results in near realtime state transitions (when it works).
* I opted to cancel the periodic timer after the service request is `started`. This means that if we're not receiving messages from the service request presence topic and the Agent ends the call, the call won't end on the Explorer side until the Explorer ends the call. Since the Explorer won't get billed for those additional seconds, I thought this was acceptable (versus keeping the timer running for potentially multiple hours). 
* If we're not receiving messages from the topic, it will take at least 3 seconds (the timer duration) to transition from `assigned` to `started`. It typically takes Dash about 2-3 seconds to [start a service request](https://github.com/aira/dashboard/blob/925aef12e02cf06a3586bf988c0b3708c711e393/src/renderer/store/modules/dashboard.js#L199-L288), so this is not out of the ordinary.
* When we removed the 90-second limit for queued service requests, we got reports of Explorers waiting for 10's of minutes without their calls being answered. Most of those cases appeared to be due to missed messages, and this change should address those too.

ipainchaud (approved):
As discussed, we know this is not ideal and that changes need to be applied in platform to insure no messages are lost. In the meantime, this will improve significantly user experience.

PR #71, opened 2023-02-25
fix: `Accept-Language` header not set on mobile

vsomayaji:
Platform uses the `Accept-Language` header to localize error messages, but the `http` package does not appear to automatically set it on mobile.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -113,7 +113,7 @@ class PlatformClient {
     Map<String, dynamic> response = await _httpPost(
       '/api/smartapp/verify/client',
       {'clientId': clientId},
-      additionalHeaders: {'Content-Type': 'application/x-www-form-urlencoded'},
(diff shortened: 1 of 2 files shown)


lib/src/platform_client.dart
```diff
@@ -917,12 +917,17 @@ class PlatformClient {
...
-      'Content-Type': 'application/json',
+      HttpHeaders.contentTypeHeader: 'application/json',
       'X-API-Key': _config.apiKey,
       'X-Client-Id': _config.clientId,
       'X-Trace-Id': traceId.toString(),
     };
 
+    if (!kIsWeb) {
+      // The http package does not automatically set the Accept-Language header on mobile.
+      headers[HttpHeaders.acceptLanguageHeader] = Platform.localeName.replaceAll('_', '-');
```
vsomayaji, line 928: `Platform.localeName` returns a string like `en_US`, but the [`Accept-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) header expects a `-`.

PR #72, opened 2023-03-06
fix: prevented a room camera switch while we are in queue

ipainchaud:
If we were in a queue and the Explorer was requesting a camera switch, the application would crash.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -253,6 +253,9 @@ class KurentoRoom extends ChangeNotifier implements Room {
 
   @override
   Future<void> startPresenting(MediaStream displayStream) async {
+    if (_localTrackId == null) {
+      return;
+    }
     // Eventually, we will create a separate connection for screen sharing. That requires changes to Platform and Dash,
     // so for now, we start presenting by replacing the Explorer's video track with the display video track.
     _presentationVideoTrack = displayStream.getVideoTracks()[0];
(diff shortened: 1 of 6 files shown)


lib/src/room.dart
```diff
@@ -274,18 +274,22 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
-    _localStream = mediaStream;
-
-    // Replace the tracks.
-    await _connectionByTrackId[_localTrackId]!.replaceAudioTrack(
-        _isAudioMuted || mediaStream.getAudioTracks().isEmpty ? null : mediaStream.getAudioTracks()[0]);
-    if (!_isPresenting) {
-      await _connectionByTrackId[_localTrackId]!
-          .replaceVideoTrack(_isVideoMuted || !_hasVideoTrack ? null : mediaStream.getVideoTracks()[0]);
-    }
+    if (_serviceRequestState == ServiceRequestState.queued) {
```
vsomayaji, line 277: This doesn't seem like the right fix, since what we really need to check for is whether the participant has joined the room. If the participant has not joined the room yet, they shouldn't be allowed to replace the stream.

How about:
```dart
if (_localStream != null) {
```

ipainchaud: I'd prefer to use `_localTrackId`

vsomayaji (approved):
A well-behaved SDK might throw a `StateError` in these scenarios instead of failing silently, but this is fine for now.

PR #73, opened 2023-03-24
feat: adds fallback procedure when data connection changes or is lost

ipainchaud:
When a user walks outside of a wifi, the application used to lose connectivity with the agent... without letting the Explorer know anything about the issue. This PR:

lets the user know when a connection is lost (not switching here, but simply lost)
without adding a reconnect button, this PR automatically attempts to reconnect as soon as a connection change happens while keeping the user aware of that pending reconnection.
uses a much more reliable reconnection process than a simple "legacy" reconnect. (we use a WebRTC restartIce).
adds a mechanism to lets the user know when a reconnection failed (This information wasn't getting back to the Explorer before)... although we still have to improve that "legacy" reconnection process which not only fails often, but doesn't provide much feedback.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:typed_data';
 
+import 'package:connectivity_plus/connectivity_plus.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_aira/src/models/position.dart';
 import 'package:flutter_webrtc/flutter_webrtc.dart';
@@ -35,6 +36,13 @@ abstract class Room implements Listenable {
 
   /// Called when a reconnect has completed.
   VoidCallback? onReconnected;
+
+  /// Called when a reconnect has failed.
+  VoidCallback? onReconnectionFailed;
+
+  /// Called the connection to Aira Servers is lost. This will happen when the device can't connect through either:
+  /// wifi, mobile data, bluetooth, ethernet or any other communication means.
+  VoidCallback? onConnectionLost;
 
   /// The ID of the service request.
   int get serviceRequestId;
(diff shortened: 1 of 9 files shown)


lib/src/room.dart
```diff
@@ -148,6 +158,32 @@ class KurentoRoom extends ChangeNotifier implements Room {
     // HACK: The `_serviceRequestPresenceTopic` has been unreliable and we haven't yet figured out why. Until then,
     // we're backing it up by periodically checking the status of the service request.
     _getServiceRequestStatusTimer = Timer.periodic(const Duration(seconds: 3), (_) => _getServiceRequestStatus());
+
+    if (defaultTargetPlatform == TargetPlatform.iOS || defaultTargetPlatform == TargetPlatform.android) {
+      _connectivityMonitoringSubscription = Connectivity().onConnectivityChanged.listen(_handleConnectionRecovery);
+    }
```
ipainchaud, line 164: Since this plugin doesn't support WEB... only register the listener when on mobile.

lib/src/room.dart
```diff
@@ -148,6 +158,32 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+          onConnectionLost?.call();
+        } else {
+          _log.info('Now using connection type: $result. '
+              'Restarting ice on ${_connectionByTrackId.length} webrtc connections!');
+          for (var element in _connectionByTrackId.values) {
+            await element.restartIce();
+          }
+          if (ConnectivityResult.none == _currentlyUsedConnectionType) {
+            await _reconnect();
+          }
```
ipainchaud, line 181: when the connection was lost (nothing to replace) and then retrieved, it seems to work better when we follow with a reconnect.

lib/src/room.dart
```diff
@@ -148,6 +158,32 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+      _connectivityMonitoringSubscription = Connectivity().onConnectivityChanged.listen(_handleConnectionRecovery);
+    }
+  }
+
+  Future<void> _handleConnectionRecovery(ConnectivityResult result) async {
+    if (result != _currentlyUsedConnectionType) { // If status changed
+      if (ServiceRequestState.started == _serviceRequestState) { // If communication was previously established
+        if (ConnectivityResult.none == result) {
+          _log.info('Lost connection to Internet');
+          onConnectionLost?.call();
```
ipainchaud, line 172: I expect that every Explorer would like to be notified if the connection is lost.

PR #75, opened 2023-05-03
feat(PlatformClient): New API to retrieve available Free Site Access.

ipainchaud:
- Added Equitable to help perform deep equality within objects
- Added Rest Endpoint access to /user/location for an easy access to available Site offer
- Added Functions to Position object to help manage Site offers and requests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Position` can be now compared using the '==' operator to know if both positions have same longitude, latitude, heading and timestamp (#75).
 - `Position.toString` will now show longitude, latitude, heading and timestamp (#75).
 - `Position.timeSinceInMs` is now an available convenience to know the amount of time elapsed since `Position.timestamp` and now (#75).
-- `Position.distanceInMetersWith` is now an available convenience to know the between two Positions (#75).
+- `Position.distanceFrom` is now an available convenience to know the between two Positions (#75).
 
 ## [1.1.1] - 2023-03-31
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.0] - 2023-04-04
+### Added
+- `RoomHandler.onAccessOfferChange` is now available to get notified of activated Site Offers (#75).
+- `PlatformClient.inquireForGPSActivatedOffer` is now available to inquire for available Site Offers when outside of a call (#75).
+- `Position` can be now compared using the '==' operator to know if both positions have same longitude, latitude and heading (#75).
+- `Position.timeSinceInMs` is now an available convenience to know the amount of time elapsed since `Position.timestamp` and now (#75).
+- `Position.distanceInMetersWith` is now an available convenience to know the between two Positions (#75).
+
 ## [1.1.1] - 2023-03-31
 ### Added
 - automatic PubNub reconnection when data connection fails or changes (#74).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `RoomHandler.onAccessOfferChange` is now available to get notified of activated Site Offers (#75).
 - `PlatformClient.inquireForGPSActivatedOffer` is now available to inquire for available Site Offers when outside of a call (#75).
-- `Position` can be now compared using the '==' operator to know if both positions have same longitude, latitude and heading (#75).
+- `Position` can be now compared using the '==' operator to know if both positions have same longitude, latitude, heading and timestamp (#75).
+- `Position.toString` will now show longitude, latitude, heading and timestamp (#75).
 - `Position.timeSinceInMs` is now an available convenience to know the amount of time elapsed since `Position.timestamp` and now (#75).
 - `Position.distanceInMetersWith` is now an available convenience to know the between two Positions (#75).
 
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -862,6 +862,10 @@ class PlatformClient {
   ///
   /// Returns null if there is no Site Offer available or an `AccessOfferDetails` of the available offer if within
   /// geofence.
+  ///
+  /// WARNING! This function is Throttled to avoid overhead on the server side. If it is called more than every seconds,
+  /// it will return the latest valid AccessOfferDetails. This is done to simplify the handling of the function as now,
+  /// we only have to worry about nulls and valid AccessOfferDetails.
   Future<AccessOfferDetails?> inquireForGPSActivatedOffer(Position position) async {
     _verifyIsLoggedIn();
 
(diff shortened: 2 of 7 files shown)


lib/src/platform_client.dart
ipainchaud, line 1: please review without whitespaces as there are many formatting changes (auto-format)

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.0] - 2023-04-04
+### Added
+- `RoomHandler.onAccessOfferChange` is now available to get notified of activated Site Offers (#75).
+- `PlatformClient.inquireForGPSActivatedOffer` is now available to inquire for available Site Offers when outside of a call (#75).
+- `Position` can be now compared using the '==' operator to know if both positions have same longitude, latitude, heading and timestamp (#75).
+- `Position.toString` will now show longitude, latitude, heading and timestamp (#75).
+- `Position.timeSinceInMs` is now an available convenience to know the amount of time elapsed since `Position.timestamp` and now (#75).
+- `Position.distanceInMetersWith` is now an available convenience to know the between two Positions (#75).
```
vsomayaji, line 14: I'm not a huge fan of the names, but no biggie.

A few comments:
*  If `onAccessOfferChange` is an optional event (similar to `onConnectionLost`), should it go on the `Room` instead of the `RoomHandler`? Does the event only apply to location-based offers or will it trigger when the Agent activates a promotion? `onAccessOfferApplied` sounds better to me.
* Instead of `inquireForGPSActivatedOffer`, why not `updateLocation` (similar to the method on the `Room`)? 
* "distance between" or "distance to/from" is more common than "distance with". And why meters? I think our app uses the Imperial system right now, correct (although we eventually want to make that a user preference)?

lib/src/platform_client.dart
```diff
@@ -837,6 +858,32 @@ class PlatformClient {
...
+  ///
+  /// Returns null if there is no Site Offer available or an `AccessOfferDetails` of the available offer if within
+  /// geofence.
+  Future<AccessOfferDetails?> inquireForGPSActivatedOffer(Position position) async {
+    _verifyIsLoggedIn();
+
+    if (null != _lastPositionUpdate && _lastPositionUpdate!.timeSinceInMs() < 1000) {
+      // Same throttling delay as `KurrentoRoom.updateLocation`.
+      return _lastAccessOfferUpdate;
+    }
```
vsomayaji, line 871: I think this throttling would make more sense if the method was named `updateLocation`. If it's an `inquire` method, it doesn't feel right to return stale/incorrect data.

lib/src/room.dart
```diff
@@ -242,6 +248,8 @@ class KurentoRoom extends ChangeNotifier implements Room {
 
   String get _serviceRequestPresenceTopic => '${_env.name}/user/${_serviceRequest.userId}/service-request/presence';
 
+  String get _triggerTopic => '${_env.name}/si/ts/${_serviceRequest.userId}';
```
vsomayaji, line 251: I don't like the name `_triggerTopic`, but I see that Platform refers to these things as "service triggers". 🤦🏾

lib/src/platform_client.dart
ipainchaud: I'll change the name.

lib/src/room.dart
ipainchaud: I did attempt to find a better name... this channel is used to trigger things. The other name I found (again inspired from legacy) is 'ToSmart' hehehe! Not amazing either. I feel we will need to change some of those topics, merge some of them as it seems like the information about the GPS activated offers should come through the same topic as the flashlight. That trigger topic has other things like the 'flashlight' in there as well.

lib/src/room.dart
```diff
@@ -442,6 +451,25 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
 
+  Future<void> _handleTriggers(String message) async {
+    Map<String, dynamic> json = jsonDecode(message);
+    _log.finest('Got following trigger: ${json['trigger']}');
+
+    // When a GPS Activated Offer is activated during a call, we get a 'SERVICE_ACCESS' message with the offer's details
+    if (json['trigger'] == 'SERVICE_ACCESS') {
+      Map<String, dynamic> value = json['value'] ?? [];
+      AccessOfferDetails accessOffer = AccessOfferDetails.fromJson(value['access']);
+      if (null == accessOffer.durationPerCall || accessOffer.durationPerCall! < 30) {
```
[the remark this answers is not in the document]
ipainchaud: I probably should have removed that. I was under the impression that we should receive a message when the offer only has 30 seconds left and I had put that there to avoid having both the "30 seconds" remaining and the initial time messages one after the other.

I'll remove it.

CHANGELOG.md
ipainchaud: about the `onAccessOfferChage`: I did it exactly how you propose it to be done. It is in `Room` but the diff is a little miss-leading because it hides some of the code.

ipainchaud: I can change the `inquireForGPSActivatedOffer` to `updateLocation`... it just sounds inaccurate to me because it doesn't update anything on the backend side.

ipainchaud: about `distanceWith`... yep! `distanceFrom` sounds a lot better.

ipainchaud: Meters because: 1. the plugins works with meters 2. the function example I've seen was in km and I got a little lazy with that I guess 3. Legacy worked with meters 4. the whole world works with meters... including Nasa 5. metric system is so much easier: it is less error prone 6. the imperial system... I mean! I thought about using it and the idea didn't fit well with my whole being 😄

lib/src/platform_client.dart
```diff
@@ -837,6 +858,32 @@ class PlatformClient {
...
+    _lastPositionUpdate = position;
+
+    String body = jsonEncode({
+      'userId': _userId,
+      'lt': position.latitude,
+      'lg': position.longitude,
+      'p': Platform.isAndroid ? 'ANDROID' : 'IOS',
+    });
+
+    Map<String, dynamic> gpsResponse = await _httpPost('/api/user/location', body);
```
pfrollo, line 881: Can you please double check that `X-Device-Id` header is set? See the platform [API](https://github.com/aira/platform/blob/269f308197299fb6ce95d27ba396a54381bdf010/core/server/AiraPlatform/src/main/java/io/aira/rest/UserRestController.java#L113).

lib/src/platform_client.dart
ipainchaud: I have added a nice comment explaining why we do the throttling. To me, `updateLocation` doesn't sound right because it doesn't update anything on the server side.

lib/src/room.dart
ipainchaud: I've renamed it to `_callEventsTopic`

CHANGELOG.md
vsomayaji: >  it just sounds inaccurate to me because it doesn't update anything on the backend side 
Are you sure? It looks like `SiteAccessManager.processGPSState` does quite a bit, including updating the `currentLocation` user property.

vsomayaji: > Legacy worked with meters

I didn't know that. When I browse location offers in the Aira - Legacy and Aira Explorer iOS apps, distances are shown in miles. Are you going to change that? If so, I'd advise against that. Most of our Explorers are in the US and the US uses miles.

In the future, allowing Explorers to choose their units makes a ton of sense, but I don't think that's a priority at the moment.

lib/src/platform_client.dart
ipainchaud: yes it does! all the _http[post, get, delete] function do by default.

PR #80, opened 2023-06-08
fix(inquireForGPSActivatedOffer): location throttling was not doing what was intended

ipainchaud:
We can't base the throttling on the Position Timestamp because this timestamp can be erroneous.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -38,7 +38,7 @@ class PlatformClient {
 
   MessagingClient? get messagingClient => _messagingClient;
 
-  Position? _lastPositionUpdate;
+  DateTime? _lastPositionUpdate;
   AccessOfferDetails? _lastAccessOfferUpdate;
 
   /// Creates a new [PlatformClient] with the specified [PlatformClientConfig].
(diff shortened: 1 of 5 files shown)


lib/src/platform_client.dart
```diff
@@ -869,11 +869,12 @@ class PlatformClient {
   Future<AccessOfferDetails?> inquireForGPSActivatedOffer(Position position) async {
     _verifyIsLoggedIn();
 
-    if (null != _lastPositionUpdate && _lastPositionUpdate!.timeSinceInMs() < 1000) {
+    DateTime now = DateTime.now();
+    if (null != _lastPositionUpdate && now.difference(_lastPositionUpdate!).inMilliseconds < 1000) {
```
pfrollo, line 873: Many people tend to walk at about 1.42 meters per second (5.1 km/h; 3.2 mph; 4.7 ft/s). Pinning platform every 1s seems excessive. I would stretch it to 5 seconds. Can we also not to call Platform when a change of distance is small?

ipainchaud: I have the tendency to believe that 1 second is a little of a small delay as well. The issue is that this API is quite a general purpose API. This can be used to refresh data in a screen, by a backend service, by a live position stream, etc. If a call to this API fails, we don't necessarily prevent it to happen again quickly. We have, upstream, other mechanisms to throttle as well. Depending on the situation, we use different throttling. [Here](https://github.com/aira/explorer-ui/blob/7e94571a71ae55f465fa6162846f74659863d0b2/lib/src/services/position_service_manager.dart#LL101C2-L101) is an example of a throttling we do when the user is in the Home Page. [Here](https://github.com/aira/explorer-ui/blob/7e94571a71ae55f465fa6162846f74659863d0b2/lib/src/services/position_service_manager.dart#L139) is an example of the throttling we do when the user's heading changes during a call (triggers a position update faster to enable the Agent to know where the Explorer points to). [Here](https://github.com/aira/explorer-ui/blob/7e94571a71ae55f465fa6162846f74659863d0b2/lib/src/services/position_service_manager.dart#L153) is an example of the distance the user must travel before another update is triggered during a call.

ipainchaud: This one second is the worst case scenario.

PR #83, opened 2023-07-03
fix(reconnect): reestablishment of presentation video track when a reconnect happens

ipainchaud:
When `Room._reconnect` was called during a presentation, the presentation stream wasn't reestablished. This is now reestablished if we were presenting.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -128,9 +128,9 @@ class KurentoRoom extends ChangeNotifier implements Room {
   bool _isVideoMuted = false;
   bool _isPresentationMuted = false;
   bool _isReconnecting = false;
-  MediaStreamTrack? _presentationVideoTrack;
+  MediaStream? _presentationStream;
 
-  bool get _isPresenting => null != _presentationVideoTrack;
+  bool get _isPresenting => null != _presentationStream;
   ServiceRequestState _serviceRequestState = ServiceRequestState.queued;
   String? _agentName;
   MediaStream? _localStream;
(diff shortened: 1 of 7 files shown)


lib/src/room.dart
```diff
@@ -308,10 +308,14 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
     // Eventually, we will create a separate connection for screen sharing. That requires changes to Platform and Dash,
     // so for now, we start presenting by replacing the Explorer's video track with the display video track.
-    _presentationVideoTrack = displayStream.getVideoTracks()[0];
-    await _connectionByTrackId[_localTrackId]!.replaceVideoTrack(_isPresentationMuted ? null : _presentationVideoTrack);
+    _presentationStream = await displayStream.clone();
+    if ((_localStream?.getAudioTracks().isNotEmpty ?? false) && _presentationStream!.getAudioTracks().isEmpty) {
+      await _presentationStream?.addTrack(_localStream!.getAudioTracks()[0]);
+    }
+    MediaStreamTrack presentationTrack = displayStream.getVideoTracks()[0];
+    await _connectionByTrackId[_localTrackId]!.replaceVideoTrack(_isPresentationMuted ? null : presentationTrack);
```
ipainchaud, line 316: We now ensure that the Presentation Stream has the audio track.

lib/src/room.dart
```diff
@@ -128,9 +128,9 @@ class KurentoRoom extends ChangeNotifier implements Room {
   bool _isVideoMuted = false;
   bool _isPresentationMuted = false;
   bool _isReconnecting = false;
-  MediaStreamTrack? _presentationVideoTrack;
+  MediaStream? _presentationStream;
```
ipainchaud, line 131: We now save the whole `MediaStream` instead of just a `MediaStreamTrack`.

lib/src/room.dart
```diff
@@ -709,7 +713,7 @@ class KurentoRoom extends ChangeNotifier implements Room {
 
     _localTrackId = track.id;
 
-    await _connectTrack(_localTrackId!, _localStream);
+    await _connectTrack(_localTrackId!, _isPresenting ? _presentationStream : _localStream);
```
ipainchaud, line 716: When we `_createOutgoingTrack`, we now create it with the right `MediaStream`.

test/flutter_aira_test.dart
```diff
@@ -70,4 +70,47 @@ void main() {
...
+      expect(speed1, equals(speed2));
+    });
+    test('Same timestamp could cause a division by zero', () {
+      var now = DateTime.now();
+      var p1 = Position(latitude: 36.00282558105645, longitude: -78.93345583177252, timestamp: now);
+      var p2 = Position(latitude: 36.00404070162735, longitude: -78.93306959369106, timestamp: now);
+      var speed = p1.speedFrom(p2);
+      expect(speed, equals(-1));
+    });
+  });
```
ipainchaud, line 115: These unit tests were among others in a future PR concerning throttling improvements. Since they really belong here I have moved them.

PR #85, opened 2023-07-17
fix(platform_client & room): use of a common location throttling mechanism to avoid concurrent updates

ipainchaud:
We have two APIs to update Explorer's locations: 

1. `Room.updateLocation` which uses a rabbitMQ API to communicate the update. This function is used to update location during a call.
2. `PlatformClient.inquireForGPSActivatedOffer` which uses a REST endpoint to know if there is a new GPS activated offer at a specific location. This function is used before calls and in the background.

These CPU intensive calls execute the same code on the backend which can cause a "endpoint modified concurrently" PlatformLocalizedException if a user ends up to execute this code while his previous call is not finished.

This PR fixes this issue by using a common throttling mechanism between these two functions and by increasing the delay between the possible calls from 1 second to 1.5 seconds.

### DIFF
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -36,8 +36,8 @@ dependencies:
 
   flutter_aira:
     path: ../
-  flutter_webrtc: ^0.9.6
-  logging: ^1.0.2
+  flutter_webrtc: ^0.9.36
+  logging: ^1.2.0
 
 dev_dependencies:
   flutter_test:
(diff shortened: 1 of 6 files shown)


example/pubspec.yaml
ipainchaud, line 1: I've updated the libraries.

pubspec.yaml
ipainchaud, line 1: I've updated the plugins here as well.

lib/src/platform_client.dart
```diff
@@ -1125,3 +1124,17 @@ class PlatformMessagingKeys {
...
+  bool get shouldThrottlePositionUpdate {
+    DateTime now = DateTime.now();
+    if (now.difference(_lastLocationUpdateTimestamp).inMilliseconds < 1500) {
+      // Same throttling delay as `KurrentoRoom.updateLocation`.
+      return true;
+    }
+    _lastLocationUpdateTimestamp = now;
+    return false;
+  }
+}
```
ipainchaud, line 1140: Behold, the fix!

chimon2000 (approved):

PR #87, opened 2023-07-27
fix(room): fixed the source of the video pixelation when starting or stoping a presentation

ipainchaud:
As reported in some of the following links:
https://airasproductt-hqa4190.slack.com/archives/C044B4WAK1U/p1689980551973879?thread_ts=1689976009.201869&cid=C044B4WAK1U
https://airasproductt-hqa4190.slack.com/archives/C044B4WAK1U/p1690404970446239
https://airasproductt-hqa4190.slack.com/archives/C044B4WAK1U/p1690417428049009

The problem:
Starting or stoping a presentation could cause a pixelation of the video on the Agent side. 

Observations:
After some experimentation, I discovered that, once I succeed to start a presentation which was causing the pixelation, the video dimensions failed to change although it should have. This "resolution" change seems to be at the source of the problem.
We also noted that a reconnect always seems to fix the issue.

The fix:
After attempting to address the "resolution" problem without success, I decided to change the way we start and end presentations for a track re-creation (which is the equivalent to a partial reconnect). 

Side effects:
The process to start and stop a presentation is longer and causes the Agent to lose the audio and video for about 2 seconds.
There is no longer any kind of pixelation.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -110,16 +110,15 @@ abstract class Room implements Listenable {
 }
 
 class KurentoRoom extends ChangeNotifier implements Room {
-
   // Private constructor.
   KurentoRoom._(
-      this._env,
-      this._client,
-      Session session,
-      this.messagingClient,
-      this._serviceRequest,
-      this._roomHandler,
-      );
+    this._env,
+    this._client,
+    Session session,
+    this.messagingClient,
+    this._serviceRequest,
+    this._roomHandler,
+  );
 
   final Logger _log = Logger('KurentoRoom');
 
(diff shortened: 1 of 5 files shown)


lib/src/room.dart
```diff
@@ -310,25 +317,63 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+          // If we are using _localStream's audio track, don't destroy it!
+          await _presentationStream!.removeTrack(_localStream!.getAudioTracks()[0]);
+        } else {
+          await _presentationStream!.getAudioTracks()[0].stop();
+        }
+      }
+      await _presentationStream!.getVideoTracks()[0].stop();
+      await _presentationStream!.dispose();
+      _presentationStream = null;
+    }
```
ipainchaud, line 363: `_presentationStream` should not contain `_localStream`'audio at this point: remove it!

lib/src/room.dart
ipainchaud, line 1: the auto-formatter has stricken!

lib/src/room.dart
```diff
@@ -310,25 +317,63 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+        if (null != _localStream &&
+            _presentationStream!.getAudioTracks()[0].id == _localStream!.getAudioTracks()[0].id) {
+          // If we are using _localStream's audio track, don't destroy it!
+          await _presentationStream!.removeTrack(_localStream!.getAudioTracks()[0]);
+        } else {
+          await _presentationStream!.getAudioTracks()[0].stop();
+        }
+      }
+      await _presentationStream!.getVideoTracks()[0].stop();
+      await _presentationStream!.dispose();
```
ipainchaud, line 361: Ok! There is some mystery here! I had some debates on if I should or should not dispose of the presentation stream here. I do have some preference in letting the creator of the stream deal with its disposition. To sooth my debates, I kept "cloning" the stream.

I haven't yet discussed about the mystery. The mystery lies in the fact that if I don't dispose of the presentation stream here, the whole operation of stoping the presentation fails with the `deleteTrack` & `createOutgoingTrack` approach. I haven't figured out why "not disposing" of the presentation stream could have those consequences.

PR #91, opened 2023-09-19
feat: Addition of terms of service

chimon2000:
This pull request adds the ability to accept the privacy policy thru the API.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -449,6 +449,24 @@ class PlatformClient {
     return User.fromJson(response);
   }
 
+  /// Updates the user's terms of service and returns
+  /// the logged-in [User].
+  Future<User> updateTermsOfService(bool tosAccepted) async {
+    _verifyIsLoggedIn();
+
+    await _httpPut(
+      '/api/user/tos',
+      body: jsonEncode(
+        {
+          'userId': _userId,
+          'tosAccepted': tosAccepted,
+        },
+      ),
+    );
+
+    return getUser();
+  }
+
   /// Returns the current billing information for the current user.
   Future<PartialBillingInformation> getPartialBillingInformation() async {
     _verifyIsLoggedIn();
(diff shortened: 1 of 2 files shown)


lib/src/platform_client.dart
```diff
@@ -449,11 +536,30 @@ class PlatformClient {
     return User.fromJson(response);
   }
 
+  /// Updates the user's terms of service and returns
+  /// the logged-in [User].
+  Future<User> updateTermsOfService(bool tosAccepted) async {
```
vsomayaji, line 541: A couple thoughts on this:
1. Whether the Terms of Service have been accepted is stored as the user property `tosAccepted`. I'm not sure why Platform exposes a separate API for updating it (i.e. `PUT /api/user/tos`). Should we continue using that API or deprecate it and just use the generic `PUT /api/user/{user_id}/property/{key}/value` API? @pfrollo?
2. Should the SDK expose a separate API for each user property or just expose the generic one (currently the private `_updatePropertyValue`) and maybe a type for user properties we care about? If we do want to stick with a separate API, this particular one should probably be named `updateTermsOfServiceAccepted`.

chimon2000: My initial thought is it generally feels a bit weird that TOS is part of the user object, it also isn't immediately obvious to the consumer that the accepted one is out of date, but I assume that is all by design.

ipainchaud: I personally like the idea of a generic function for all property values with the use of an enum. This wasn't implemented that way because we had only one property. I would love to avoid multiplying code like when we have an elegant solution.

pfrollo: My apology. I missed this question as there were several emails regarding this flutter PR and I don't pay much attention to Flutter. I will look into this first thing tomorrow morning.

pfrollo: `PUT /api/user/tos` updates `TOS` using `userId` and a boolean `tosAccepted`.

The `PUT /api/user/{user_id}/property/{key}/value` deletes all user properties and inserts new properties. I don't think this is the API to be called in this case.

PR #92, opened 2023-09-27
feat: Added a way to communicate with the user that a connection has failed

ipainchaud:
Before, if a sfu connection failed, there was no way for the user of the api to know about it. This fix adds a callback for this event to be communicated. Since that callback is not mandatory to provide, it is not a breaking change.


lib/src/platform_client.dart
ipainchaud, line 1: nothing new here.

lib/src/room.dart
```diff
@@ -43,6 +43,9 @@ abstract class Room implements Listenable {
   /// wifi, mobile data, bluetooth, ethernet or any other communication means.
   VoidCallback? onConnectionLost;
 
+  /// Called when a participant to the call has failed to connect.
+  VoidCallback? onConnectionFailed;
```
ipainchaud, line 47: Here is the callback. I decided against using a new state because we might have more than one participant in the future and I didn't want to risk the error status to be nuked by another status.

lib/src/room.dart
```diff
@@ -666,6 +672,9 @@ class KurentoRoom extends ChangeNotifier implements Room {
   void _handleConnectionState(int trackId, RTCPeerConnectionState state) {
     // REVIEW: Should the room expose the connection state (e.g. `isAgentStreamConnected`, `isExplorerStreamConnected`)?
     _log.info('connection state changed track_id=$trackId state=$state');
+    if (RTCPeerConnectionState.RTCPeerConnectionStateFailed == state) {
+      onConnectionFailed?.call();
+    }
```
ipainchaud, line 677: We call that callback here... if not null.

lib/src/sfu_connection.dart
```diff
@@ -30,6 +30,9 @@ class SfuConnection {
   late final RTCPeerConnection _peerConnection;
   late final RTCRtpTransceiver _audio;
   late final RTCRtpTransceiver _video;
+  RTCPeerConnectionState? _connectionState;
+
+  bool get isConnectionFailed => _connectionState == RTCPeerConnectionState.RTCPeerConnectionStateFailed;
```
ipainchaud, line 35: These are not used yet, but I can see myself want to know which SFU connection has failed.

lib/src/sfu_connection.dart
```diff
@@ -45,12 +48,14 @@ class SfuConnection {
...
-        _getConfiguration(stunServers, turnServers);
+    Map<String, dynamic> configuration = _getConfiguration(stunServers, turnServers);
 
     // Create the peer connection.
     _peerConnection = await createPeerConnection(configuration)
-      ..onConnectionState = ((state) => onConnectionState.call(trackId, state))
+      ..onConnectionState = ((state) {
+        _connectionState = state;
+        onConnectionState.call(trackId, state);
+      })
```
ipainchaud, line 58: Setting the status of the SFU connection here.

pubspec.yaml
ipainchaud, line 1: some library upgrades.

lib/src/room.dart
```diff
@@ -666,6 +673,10 @@ class KurentoRoom extends ChangeNotifier implements Room {
   void _handleConnectionState(int trackId, RTCPeerConnectionState state) {
     // REVIEW: Should the room expose the connection state (e.g. `isAgentStreamConnected`, `isExplorerStreamConnected`)?
     _log.info('connection state changed track_id=$trackId state=$state');
+    if (RTCPeerConnectionState.RTCPeerConnectionStateFailed == state) {
```
vsomayaji, line 676: If the connection is in the `failed` state, why do we remove it and call `onConnectionFailed` instead of calling `restartIce` like we do when connectivity changes?

ipainchaud: Good idea! I'll add a retry mechanism.

ipainchaud: I have checked a couple of things here and `restartIce` would have to be called on the `RTCPeerConnection` object -- which has failed. I could recreate the full "peerConnection", but this would not be a trivial task ask the SFUConnection's creation is async and will succeed or fails way after the `createPeerConnection` (in `SFUConnection.connect`) returns we our state at that time can vary.

All this combined with the fact that this error will happen pretty shortly after the Explorer attempts to call Aira seems to indicate this would not be too much to ask our Explorer to retry.

All that would be missing is the added simplicity to be returned to the "call with message" page for an easy retry without loosing message or file to share with the agent. This is what I opted to do to remedy to that issue and you should be able to see an additional commit to PR1080 addressing this.

lib/src/room.dart
```diff
@@ -666,6 +673,10 @@ class KurentoRoom extends ChangeNotifier implements Room {
...
+      SfuConnectionDirection direction = _connectionByTrackId.remove(trackId)!.direction;
```
vsomayaji, line 677: Do we need to dispose the removed connection?

vsomayaji (approved):

lib/src/room.dart
ipainchaud: good question! I'll call the method just in case.

vsomayaji (approved):
:shipit:

PR #94, opened 2023-10-05
feat: added endpoint to delete account

ipainchaud:
Simple use of rest delete on the right endpoint.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -24,8 +24,7 @@ class PlatformClient {
   ///
   /// [httpClient] can be provided if you want to use your own HTTP client (e.g. a
   /// [`SentryHttpClient`](https://docs.sentry.io/platforms/dart/usage/advanced-usage/)).
-  PlatformClient(this._config, [http.Client? httpClient])
-      : _httpClient = httpClient ?? http.Client();
+  PlatformClient(this._config, [http.Client? httpClient]) : _httpClient = httpClient ?? http.Client();
 
   final _log = Logger('PlatformClient');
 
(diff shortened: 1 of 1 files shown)


lib/src/platform_client.dart
```diff
@@ -88,8 +86,7 @@ class PlatformClient {
       'phoneNumber': phoneNumber,
     });
 
-    Map<String, dynamic> response =
-        await _httpPost('/api/smartapp/verify/confirm', body);
+    Map<String, dynamic> response = await _httpPost('/api/smartapp/verify/confirm', body);
```
pfrollo, line 89: Did you know that in IDEA you can select to reformat only the touched lines to avoid large PRs? 
Screenshot 2023-10-05 at 11 43 42 AM

lib/src/platform_client.dart
```diff
@@ -247,6 +240,10 @@ class PlatformClient {
     return loginWithCredentials(credentials);
   }
 
+  /// Delete an account.
+  /// This action is irreversible and lead to permanent deletion of all data within the account.
+  Future<void> deleteAccount() => _httpDelete('/api/user/$_userId}');
```
chimon2000, line 245: question: Is the bracket at the end of this desired?

chimon2000 (approved):
+1 this PR mostly looks good but I had one question

lib/src/platform_client.dart
ipainchaud: ????!!! nop! Thanks!

lib/src/platform_client.dart
ipainchaud: I had forgotten about that one! thanks!

PR #95, opened 2023-10-06
feat: add `PlatformBusinessLoginRequiredException`

vsomayaji:
This exception is thrown by Platform when an operation requires the user to log in with their business credentials. The name of the required Auth0 enterprise connection is available in the `connection` property.


ipainchaud (approved):
I should have added this to the changelog... I'll add it

PR #96, opened 2023-10-16
feat: new Exception to indicate account deletion failure reasons

ipainchaud:
`PlatformClient.deleteAccount` is now capable to throw `PlatformDeleteAccountException` to indicate the failure reason.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -124,7 +124,7 @@ class KurentoRoom extends ChangeNotifier implements Room {
     this._roomHandler,
   );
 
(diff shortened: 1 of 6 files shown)


lib/src/room.dart
```diff
@@ -167,6 +167,7 @@ class KurentoRoom extends ChangeNotifier implements Room {
       await room._init(session);
       return room;
     } catch (e) {
+      _log.shout('Unable to initialize the WebRTC Room.', e);
```
ipainchaud, line 170: This should help us greatly to debug issues in the future. Let's hope this is not too frequent.

PR #97, opened 2023-11-06
fix(location): Increased the throttling duration to 2 seconds for location updates

ipainchaud:
Since we are still getting a good amount of sentry error like the following:

> PlatformLocalizedException: The endpoint has been modified concurrently, and it has an exclusive lock. Please try again later

We have moved the maximum of location update request to one every 2 seconds.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -10,6 +10,7 @@ import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter_aira/flutter_aira.dart';
 import 'package:flutter_aira/src/messaging_client.dart';
 import 'package:flutter_aira/src/models/sent_file_info.dart';
+import 'package:flutter_aira/src/throttler.dart';
 import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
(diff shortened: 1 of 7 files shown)


lib/src/platform_client.dart
```diff
@@ -1330,15 +1331,7 @@ class PlatformMessagingKeys {
...
-  DateTime get lastLocationUpdateTimestamp => _lastLocationUpdateTimestamp;
-  bool get shouldThrottlePositionUpdate {
-    DateTime now = DateTime.now();
-    if (now.difference(_lastLocationUpdateTimestamp).inMilliseconds < 1500) {
-      // Same throttling delay as `KurrentoRoom.updateLocation`.
-      return true;
-    }
-    _lastLocationUpdateTimestamp = now;
-    return false;
-  }
```
ipainchaud, line 1343: Since the error was suggesting that we might have an issue with our throttling algorithm, I moved that code in a class for easier re-use and testability.

test/throttler_test.dart
ipainchaud, line 1: Added these tests to ensure our throttling really behaves the way we are expecting it to.

PR #102, opened 2024-01-04
feat: add `User.aiDailyMessageLimit`

evgeniya-aira:
We are adding a user property `aiDailyMessageLimit` that will determine how many AI image description requests an Explorer can make per day.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.11.0] - 2024-01-03
 ### New
 - `User.aiDailyMessageLimit` to determine how many AI image description requests an Explorer can make per day (#102)
-- `User.getAIDailyMessageLimitProperty` to retrieve property from `UserProperties` (#102)
 
 ## [1.10.1] - 2023-12-28
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.11.0] - 2024-01-03
+### New
+- `User.aiDailyMessageLimit` to determine how many AI image description requests an Explorer can make per day (#102)
+- `User.getAIDailyMessageLimitProperty` to retrieve property from `UserProperties` (#102)
+
 ## [1.10.1] - 2023-12-28
 ### Changed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.11.0] - 2024-01-03
+### New
+- `User.aiDailyMessageLimit` to determine how many AI image description requests an Explorer can make per day (#102)
+- `User.getAIDailyMessageLimitProperty` to retrieve property from `UserProperties` (#102)
```
vsomayaji, line 10: This is a private method, so you don't need to mention it in the changelog.

lib/src/models/user.dart
```diff
@@ -11,6 +11,7 @@ class User {
   final List<Profile> profiles;
   final String? referralLink;
   final bool tosAccepted;
+  final String? aiDailyMessageLimit;
```
vsomayaji, line 14: Thinking about this more, how about we make this an `int` instead of a `String?`? We can do the conversion here so apps don't have to deal with the `null` and `NaN` scenarios themselves.

Maybe something like:
```dart
int _getAIDailyMessageLimitProperty(Map<String, dynamic> json) =>
    int.tryParse(json['aiDailyMessageLimit']?.first?['value'] ?? '0') ?? 0;
```

CHANGELOG.md
evgeniya-aira: Okay, I'll remove it

lib/src/models/user.dart
```diff
@@ -11,6 +11,7 @@ class User {
   final List<Profile> profiles;
   final String? referralLink;
   final bool tosAccepted;
+  final int? aiDailyMessageLimit;
```
vsomayaji, line 14: Let's make this an `int` instead of an `int?`. `_getAIDailyMessageLimitProperty` won't return `null`.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.11.0] - 2024-01-03
+### New
```
vsomayaji, line 8: Not critical, but if we're following [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), this should be `Added` instead of `New`.

vsomayaji (approved):
:shipit:

PR #103, opened 2024-01-11
sendExplorerFeedback add in SDK 

evgeniya-aira:
We are adding `sendMessageFeedback` to provide functionality to allow users to leave feedback on AI response messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.12.0] - 2024-01-10
+### Added
+- `ChatMessageInfo.ChatFeedbackInfo` to include Explorer feedback (#107)
+- `Message.messageId` (#107)
+- `PlatformClient.sendMessageFeedback` to send user feedback to the AI response (#107)
+
 ## [1.11.0] - 2024-01-03
 ### Added
-- `User.aiDailyMessageLimit` to determine how many AI image description requests an Explorer can make per day (#102)
+- `User.aiDailyMessageLimit` to determine how many AI image description requests an Explorer can make per day (#106)
 
 ## [1.10.1] - 2023-12-28
 ### Changed
--- a/lib/flutter_aira.dart
+++ b/lib/flutter_aira.dart
@@ -5,6 +5,7 @@ export 'src/models/account.dart' show Account, AccountType;
 export 'src/models/billing_info.dart' show PartialBillingInformation;
 export 'src/models/call_history.dart' show CallSession, SessionFeedback;
 export 'src/models/chat.dart' show ChatMessageInfo, ChatSessionInfo, SenderRole;
+export 'src/models/chat_feedback_info.dart' show ChatFeedbackInfo;
 export 'src/models/credentials.dart' show Credentials;
 export 'src/models/feedback.dart' show AgentFeedback, Feedback, Rating;
 export 'src/models/message.dart' show Message;
--- a/lib/src/messaging_client.dart
+++ b/lib/src/messaging_client.dart
@@ -102,6 +102,7 @@ content={message: {senderId: 6187, serviceId: 88697, text: with one picture}, fi
         sentAt: envelope.publishedAt.toDateTime().millisecondsSinceEpoch,
         text: content['text'] ?? '',
         userId: content['senderId'],
+        messageId: content['messageId'] ?? int.parse(envelope.uuid.value),
         fileId: fileInfo['id'],
         fileName: fileInfo['name'],
       );
--- a/lib/src/models/chat_feedback_info.dart
+++ b/lib/src/models/chat_feedback_info.dart
@@ -1,0 +1,15 @@
+class ChatFeedbackInfo {
+  final String authorId;
+  final int rating;
+  final String? comment;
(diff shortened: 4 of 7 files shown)


lib/src/platform_client.dart
```diff
@@ -1089,6 +1089,19 @@ class PlatformClient {
     return ChatMessageInfo.fromJson(response);
   }
 
+  /// Sends user feedback on the AI response.
+  Future<void> sendMessageFeedback(int sessionId, int messageId, int rating, String comment) async {
```
vsomayaji, line 1093: For consistency with the other chat APIs, let's rename the method to `sendChatMessageFeedback` (to distinguish from other "messages") and rename `sessionId` to `chatId`.

The other feedback method in this class is `saveFeedback`. I'm not sure if I like `save` or `send` (or something else) better, so I'm fine leaving it `send` for now.

lib/src/models/chat_feedback_info.dart
```diff
@@ -0,0 +1,15 @@
+class ChatFeedbackInfo {
```
vsomayaji, line 1: All of the other chat-related models are currently in `lib/src/models/chat.dart`, so how about we move this there to keep them together?

lib/src/models/chat_feedback_info.dart
```diff
@@ -0,0 +1,15 @@
...
+  final String authorId;
+  final int rating;
+  final String? comment;
+
+  ChatFeedbackInfo({required this.authorId, required this.rating, this.comment});
+
+  factory ChatFeedbackInfo.fromJson(Map<String, dynamic> json) {
```
vsomayaji, line 8: In other places, we use a named constructor instead of a factory. Why use a factory here?

```dart
  ChatFeedbackInfo.fromJson(Map<String, dynamic> json)
      : authorId = json['authorId'],
        rating = json['rating'],
        comment = json['comment'];
```

https://medium.com/nerd-for-tech/named-constructor-vs-factory-constructor-in-dart-ba28250b2747

lib/src/messaging_client.dart
```diff
@@ -102,6 +102,7 @@ content={message: {senderId: 6187, serviceId: 88697, text: with one picture}, fi
         sentAt: envelope.publishedAt.toDateTime().millisecondsSinceEpoch,
         text: content['text'] ?? '',
         userId: content['senderId'],
+        messageId: content['messageId'] ?? int.parse(envelope.uuid.value),
```
vsomayaji, line 105: Hmm. Converting the UUID to an int here worries me, as that seems like something that could fail.

What do you think about changing `messageId` from an `int` to a `int?`? Since we only need the value for AI chat messages at the moment, maybe that's safe?

CHANGELOG.md
```diff
@@ -4,9 +4,15 @@ All notable changes to this project will be documented in this file.
...
+## [1.12.0] - 2024-01-10
+### Added
+- `ChatMessageInfo.ChatFeedbackInfo` to include Explorer feedback (#107)
+- `Message.messageId` (#107)
+- `PlatformClient.sendMessageFeedback` to send user feedback to the AI response (#107)
+
 ## [1.11.0] - 2024-01-03
 ### Added
-- `User.aiDailyMessageLimit` to determine how many AI image description requests an Explorer can make per day (#102)
+- `User.aiDailyMessageLimit` to determine how many AI image description requests an Explorer can make per day (#106)
```
vsomayaji, line 15: Why change this from #102?

CHANGELOG.md
```diff
@@ -4,9 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.12.0] - 2024-01-10
+### Added
+- `ChatMessageInfo.ChatFeedbackInfo` to include Explorer feedback (#107)
```
vsomayaji, line 9: This PR's number is #103.

lib/src/models/chat_feedback_info.dart
evgeniya-aira: Okay, I'll move it

lib/src/models/chat_feedback_info.dart
evgeniya-aira: Make sense. Thank you

CHANGELOG.md
evgeniya-aira: No, but the number was't correct in the previous commit, so I have changed to prevent future misunderstanding

lib/flutter_aira.dart
```diff
@@ -4,7 +4,8 @@ export 'src/models/access_offer.dart' show AccessOfferDetails, AccessOfferType,
 export 'src/models/account.dart' show Account, AccountType;
 export 'src/models/billing_info.dart' show PartialBillingInformation;
 export 'src/models/call_history.dart' show CallSession, SessionFeedback;
-export 'src/models/chat.dart' show ChatMessageInfo, ChatSessionInfo, SenderRole;
+export 'src/models/chat/chat.dart' show ChatMessageInfo, ChatSessionInfo, SenderRole;
+export 'src/models/chat/chat_feedback_info.dart' show ChatFeedbackInfo;
```
vsomayaji, line 8: I meant that we could include `ChatFeedbackInfo` in the existing `chat.dart` since it contains the three other chat-related exports; however, this is fine.

vsomayaji (approved):

vsomayaji (approved):

PR #105, opened 2024-01-31
fix: improved error logging from rabbit MQ

ipainchaud:
Some messages received from Platform are not encoded properly and this results in sentry errors.
For more information see https://github.com/aira/platform/pull/1111


ipainchaud:
I got some small privacy concerns in displaying message contents when there is an error. Closing this PR.

PR #107, opened 2024-02-14
New get request was added to the plugin 

evgeniya-aira:
`getChatSession` was added to the plugin in order to list of messages form the platform

### DIFF
--- a/lib/src/models/chat_feedback_info.dart
+++ b/lib/src/models/chat_feedback_info.dart
@@ -1,0 +1,15 @@
+class ChatFeedbackInfo {
+  final String authorId;
+  final int rating;
+  final String? comment;
+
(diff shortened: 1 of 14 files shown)


kasemj (approved):
LGTM

lib/src/models/chat_feedback_info.dart
```diff
@@ -0,0 +1,15 @@
+class ChatFeedbackInfo {
+  final String authorId;
+  final int rating;
+  final String? comment;
+
+  ChatFeedbackInfo({required this.authorId, required this.rating, this.comment});
```
ipainchaud, line 6: quibble: we usually follow Dart and Flutter's code style which puts constructors before properties.

PR #118, opened 2024-03-07
chore: updated MQTT's WebSocket protocol for better support of newer RabbitMQ versions.

ipainchaud:
Our effort to keep RabbitMQ up to date brought an incompatibility issue: 

> Maybe in the rabbitmq_web_mqtt v3.12.13 plugin there was a change on how incoming connections are handled

This PR changes that protocol and was tested with both the new version of RabbitMQ and the previous one to ensure backward compatibility


KasemJaffer (approved):

vsomayaji (approved):
:shipit:

PR #119, opened 2024-03-08
Left event support and change default in ServiceRequestState

smunozp:
Handle Left event triggered when there is a call transfer and Agent 1 leaves the call


ipainchaud:
I'm still waiting on the team to confirm that this is the best design we could use, but don't want to paralyze the Call Transfer efforts. So I'll create a release with this change and mark it as being under test in the release notes. If this goes well, we will only have to update the release notes. If we we don't want this change, we will need to delete the release.

ipainchaud (approved):

PR #120, opened 2024-04-11
Update the enum class [AgentFeedbackState]

kasemj:
This PR 

1. Updates the enum class `AgentFeedbackState` to include the new state `AgentFeedbackState.assigned`
2. Replaces `AgentFeedbackState.inProgress` with `AgentFeedbackState.started`

Full list can be found [here](https://github.com/aira/aira-backend-services/blob/main/api-models/src/main/resources/openapi/task-manager.yaml#L693-L697)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.0] - 2024-04-11
+
+### Changed
+- Updated the enum class [AgentFeedbackState] to include the new state [AgentFeedbackState.assigned] and replaced [AgentFeedbackState.inProgress] with [AgentFeedbackState.started] (#120)
+
 ## [2.2.0] - 2024-03-08
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.3.0] - 2024-04-11
 
 ### Changed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.0] - 2024-04-11
+
+### Changed
+- Updated the enum class [AgentFeedbackState] to include the new state [AgentFeedbackState.assigned] and replaced [AgentFeedbackState.inProgress] with [AgentFeedbackState.started] (#120)
```
ipainchaud, line 10: This has to be a breaking change. You must prefix this change log with \*\*BREAKing\*\*

lib/src/models/chat/agent_feedback_info.dart
```diff
@@ -78,17 +78,23 @@ class AgentFeedbackInfo {
...
   const AgentFeedbackState(this.value);
 
   /// Gets the [AgentFeedbackState] from the given value.
-  static AgentFeedbackState? fromValue(String? value) {
-    return AgentFeedbackState.values.firstWhereOrNull((e) => e.value == value);
+  ///
+  /// If the value is not found, [pending] is returned.
+  static AgentFeedbackState fromValue(String? value) {
+    return AgentFeedbackState.values
+            .firstWhereOrNull((e) => e.value == value) ??
```
ipainchaud, line 97: why not `AgentFeedbackState.value.firstWhere((e) => e.value == value, orElse: () => pending)` ???

ipainchaud (changes requested):

lib/src/models/chat/agent_feedback_info.dart
KasemJaffer: I agree, it doesn't look as cleaner now as it used when i was returning null for not found. I will change it just to make you satisfied :D

kasemj: I agree, it doesn't look as cleaner now as it used when i was returning null for not found. I will change it just to make you satisfied :D

ipainchaud (approved):
Looks Good!

PR #121, opened 2024-04-16
fix: addressing the 'Unable to RTCPeerConnection::addCandidate: Error The remote description was null' issue

ipainchaud:
WebRTC depends on a series of events. The SDP offer is a step which must happen before we start registering ICE Candidates (routes). This fix basically stores potential ICE Candidates until we have a SDP offer accepted. Then it registers the stored Candidates and allow subsequents offer to be added directly.

https://linear.app/airaio/issue/EXPO-346/sentry-the-remote-description-was-null


pubspec.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
   logging: ^1.2.0
   meta: ^1.10.0
   mqtt_client: ^10.2.0
+  mutex: ^3.1.0
```
bwang-aira, line 24: I think your approach of storing ICE candidates before remote description becomes available, is generally a good approach. I'm questioning the usage of `mutex` here. Dart code here is single-threaded anyway, so racing condition cannot occur, so I wonder if you have any good reasons to use mutex, and if not using mutex, were you seeing any issues?

ipainchaud: Yeah! I did start without that Mutex protection for the same reasons you are mentioning. But then I got the surprise of having concurrency issue with the call to `_peerConnection.addCandidate(candidate)` as without that mutext we end up calling that function concurrently (call 1 is not finished and we perform call 2).

So, in other words, this protection is not so much to protect the access to the list of candidates, but more to protect agains having multiple concurrent calls to `addCandidate`.

ipainchaud: I'll add a comment to clarify this.

bwang-aira (dismissed):

PR #125, opened 2024-06-19
Feat/access ai preferences

HoseinHaqiqian:
Added functions (`updateAiLanguageLevel`, `updateAiVerbosity`) and properties to `User` model.

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -584,6 +584,21 @@ class PlatformClient {
     );
   }
 
+  /// Used to update the preferred ai language level of a user.
+  Future<void> updateAiLanguageLevel(AiLanguageLevel languageLevel) async {
+    await setUserProperty(
+      UserProperty.aiLanguageLevel,
+      languageLevel.toValue(),
+    );
+  }
+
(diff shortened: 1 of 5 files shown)


lib/src/platform_client.dart
```diff
@@ -584,6 +584,21 @@ class PlatformClient {
     );
   }
 
+  /// Used to update the preferred ai language level of a user.
+  Future<void> updateAiLanguageLevel(AiLanguageLevel languageLevel) async {
+    await setUserProperty(
```
kasemj, line 589: Not a big deal but I would recommend returning the future itself instead of waiting for it inside the function. Since you're not doing anything after the await. 

```dart
Future<void> updateAiLanguageLevel(AiLanguageLevel languageLevel) {
    return setUserProperty(
```

kasemj (approved):
LGTM

lib/src/platform_client.dart
HoseinHaqiqian: Actually, i didn't pay enough attention to this matter as i just copied existing similar functions and go : )). but i agree 100%

PR #135, opened 2024-08-17
Allow server to specify the ice servers configuration

fgutmann:
Currently, the final ice servers used for a peer connection are determined by a list of stun and turn servers provided by platform, which are then conveted into a list of ice servers on the client.

The current setup gives the server limited control over how the final ice servers are calculated.

This change allows platform to return a field "iceServers", which take presedence over the stunServer and turnServer fields, if specified. When the field is not populated by platform, the app continues to use the old logic.

This change allows us to experiment with different turn services such as cloudflare's turn service or registering tcp URLs for the turn server by changing server side logic only.

### DIFF
--- a/lib/src/models/service_request.dart
+++ b/lib/src/models/service_request.dart
@@ -19,6 +19,9 @@ class ServiceRequest {
   final int userId;
   final List<dynamic> stunServers;
   final List<dynamic> turnServers;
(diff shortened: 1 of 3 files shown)


kasemj (approved):
LGTM

HoseinHaqiqian (approved):
LGTM

PR #136, opened 2024-08-19
Multiple agent support

HoseinHaqiqian:
This adds feature to manage multiple agents tracks while doing Transfer call, including:

- Adds agents track to memory by agentId while agent joins.
- Removes the agent track when an agent leaves the call.

#### Conclusion: 
This makes Explorer or ASL app to have control over tracks. resulting in less memory usage than before as well as switching between agents video track when an agent leaves during the transfer call.

### DIFF
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -18,7 +18,7 @@ publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 version: 1.0.0+1
 
 environment:
-  sdk: ">=2.16.2 <3.0.0"
+  sdk: ">3.0.0"
 
 # Dependencies specify other packages that your package needs in order to work.
 # To automatically upgrade your package dependencies to the latest versions
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -400,6 +400,7 @@ class PlatformClient {
           .split(' ')
           .first, // Split the first name and last initial.
       'status': response['serviceStatus'],
+      'agentId': response['agentId'].toString()
     };
   }
 
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -1100,10 +1100,14 @@ class KurentoRoom extends ChangeNotifier implements Room {
       for (SfuConnection connection in _connectionByTrackId.values) {
         await connection.dispose();
       }
+
+      _connectionByTrackId.keys.forEach((trackId) async {
+        await _roomHandler.removeRemoteStream(trackId);
+      });
+
       _connectionByTrackId.clear();
       _incomingTrackIdByOutgoingTrackId.clear();
       _tracksByAgentId.clear();
-
       // Create and connect new tracks to receive the remote streams from the other participant(s). (Do this before our
       // local stream, since we want to be able to hear the Agent ASAP.)
       for (Participant participant
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -157,7 +157,7 @@ class KurentoRoom extends ChangeNotifier implements Room {
 
   bool get _isPresenting => null != _presentationStream;
   ServiceRequestState _serviceRequestState = ServiceRequestState.queued;
-  final Map<String, String> _agentsNames = {};
+  final Map<String, String> _agentsName = {};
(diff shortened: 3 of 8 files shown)


lib/src/room.dart
```diff
@@ -919,6 +938,13 @@ class KurentoRoom extends ChangeNotifier implements Room {
 
   Future<void> _handleTrack(int trackId, RTCTrackEvent event) async {
     await _roomHandler.addRemoteStream(trackId, event.streams[0]);
+    // adds latest incoming tracks to the last agent
+    final currentAgent = agentsNames.keys.last;
+    if (_tracksByAgentId[currentAgent] == null) {
+      _tracksByAgentId[currentAgent] = [trackId];
+    } else {
+      _tracksByAgentId[currentAgent]!.add(trackId);
+    }
```
ming-chu, line 947:
```suggestion
    _tracksByAgentId.putIfAbsent(currentAgent, () => []).add(trackId);
```

lib/src/platform_client.dart
```diff
@@ -400,6 +400,7 @@ class PlatformClient {
           .split(' ')
           .first, // Split the first name and last initial.
       'status': response['serviceStatus'],
+      'agentId': response['agentId'].toString()
```
ming-chu, line 403: I assumed the `agentId` always a non-null value and this should be fine.

lib/src/room.dart
```diff
@@ -1107,4 +1133,10 @@ class KurentoRoom extends ChangeNotifier implements Room {
       _isReconnecting = false;
     }
   }
+
+  void _removeTracksByAgent(int incomingTrackId) {
+    _tracksByAgentId.removeWhere((key, value) {
+      return value.firstWhereOrNull((item) => item == incomingTrackId) != null;
```
ming-chu, line 1139: I guess we may simplify this line by use any():
```suggestion
      return value.any((item) => item == incomingTrackId);
```

example/pubspec.yaml
```diff
@@ -18,7 +18,7 @@ publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 version: 1.0.0+1
 
 environment:
-  sdk: ">=2.16.2 <3.0.0"
+  sdk: ">3.0.0"
```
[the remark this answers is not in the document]
HoseinHaqiqian: This is just for example project of flutter_aira. not the plugin itself. so It's not really sensitive.

lib/src/platform_client.dart
HoseinHaqiqian: `agentId` is Int, trying to make it String for further use.

lib/src/room.dart
HoseinHaqiqian: Ahh, I missed this. how about `contains`? its simpler.

ming-chu: yes, if you don't need to have a custom `test` function and this case use `contains` could be even better.

ming-chu (approved):
LGTM

lib/src/room.dart
```diff
@@ -793,7 +807,11 @@ class KurentoRoom extends ChangeNotifier implements Room {
         break;
       // Handles the case when the Agent leaves the room on a call transfer, name is turned to null to represent that agent is no longer in the call
       case 'LEFT':
-        _agentName = null;
+        _agentsName.remove(agentId.toString());
+        final tracks = _tracksByAgentId[agentId];
+        tracks?.forEach((trackId) async {
```
kasemj, line 812: Nit picking here. But I would say to prefer `for` loop over `forEach` functions for several reasons but especially when using async closures. One thing to note in this case is that `forEach` doesn't care about what the closure returns so if you return future it will not wait for it. Might not be relevant if you don't care about waiting for each item to finish executing.

HoseinHaqiqian: First I didn't care about being sequence sensitive but know that you mentioned this I would like to be more cautions on this.

kasemj: Thanks!

kasemj (dismissed):

kasemj (approved):

PR #137, opened 2024-08-19
Limit access ai

evgeniya-aira:
Added a new accessAi property to the User's Account (Profile) data.

**Details**
Introduced the accessAi property to the User's Account (Profile) model.
This property is intended to manage Access AI permissions to different user's profiles.

**Example**:
A response example including the new `accessAi` property can be found in the following [issue](https://linear.app/airaio/issue/PLAT-269/limit-access-ai-platform-add-support-for-explorer-app#comment-b851829b).

### DIFF
--- a/lib/src/models/account.dart
+++ b/lib/src/models/account.dart
@@ -15,13 +15,35 @@ enum AccountType {
   }
 }
 
+enum AccessAi {
+  fullAccess('FULL_ACCESS'),
+  noHistory('NO_HISTORY'),
+  noAccess('NO_ACCESS');
+
+  final String value;
+
+  const AccessAi(this.value);
+
+  /// Gets the [AccessAi] from the given value.
+  ///
+  /// If the value is not found, [noAccess] is returned.
+  static AccessAi fromValue(String? value) {
+    return AccessAi.values
+        .firstWhere((e) => e.value == value, orElse: () => AccessAi.noAccess);
+  }
+}
+
 class Account {
   final int id;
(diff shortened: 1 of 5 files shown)


lib/src/models/account.dart
```diff
@@ -15,13 +15,35 @@ enum AccountType {
...
+  final String value;
+
+  const AccessAi(this.value);
+
+  /// Gets the [AccessAi] from the given value.
+  ///
+  /// If the value is not found, [noAccess] is returned.
+  static AccessAi fromValue(String? value) {
+    return AccessAi.values
+        .firstWhere((e) => e.value == value, orElse: () => AccessAi.noAccess);
```
ming-chu, line 32: I am not sure if we want to make it upper case `.toUpperCase()` before compare.

ming-chu (approved):
LGTM

lib/src/models/account.dart
ming-chu: image

if the [API](https://linear.app/airaio/issue/PLAT-269/limit-access-ai-platform-add-support-for-explorer-app#comment-b851829b) always return upper case, it should be fine.

evgeniya-aira: Screenshot 2024-08-20 at 9 14 02 AM
It looks like it's in uppercase in the [API](https://github.com/aira/platform/blob/master/core/server/AiraPlatformData/src/main/java/io/aira/data/model/Account.java)

PR #140, opened 2024-09-11
Fix API exception parser

ming-chu:
Only return `PlatformLocalizedException` when `errorMessage` is not empty.

existing behavior:
when `errorMessage` from API response is empty string, it will throw `PlatformLocalizedException`.

expected behavior:
when `errorMessage` from API response is empty or null, it should throw `PlatformUnknownException`.

try to handle response from "https://dev-platform.aira.io/api/user/location" has empty(non-null) errorMessage when it's not `SUCCESS`.
image

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -1503,7 +1503,7 @@ class PlatformClient {
     if (errorCode == 'SEC-001') {
       return const PlatformInvalidTokenException();
     }
-    final errorMessage = json['response']?['errorMessage'];
+    final errorMessage = json['response']?['errorMessage'] as String? ?? '';
     final metadataConnection = json['metadata']?['connection'];
     if (errorCode == 'AIRA-ACCESS-017' && metadataConnection != null) {
       return PlatformBusinessLoginRequiredException(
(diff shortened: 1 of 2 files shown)


lib/src/platform_client.dart
```diff
@@ -1480,6 +1462,36 @@ class PlatformClient {
           MessagingClientPubNub(_config.messagingKeys!, _userId, token);
     }
   }
+
+  @visibleForTesting
+  static Exception? getExceptionFromApiResponse({
```
HoseinHaqiqian, line 1467:
```suggestion
// separate class
enum ErrorCodes {
  invalidToken('SEC-001'),
  businessLoginRequired('AIRA-ACCESS-017'),
  deleteAccount('KN-UM-065');

  const ErrorCodes(this.code);

  final String code;
}
// in getExceptionFromApiResponse method

switch (ErrorCodes.fromCode(errorCode)) {
      case ErrorCodes.invalidToken:
        return const PlatformInvalidTokenException();
      case ErrorCodes.businessLoginRequired:
      // TODO: Handle this case.
      case ErrorCodes.deleteAccount:
      // TODO: Handle this case.
    }


  
  
  
```

ming-chu: Nice! I like this! 👍 
I will update it If we going to accept this PR. Since the original issue is not happen any more.

ming-chu:
Close it now, since the issue is not happen anymore and we might want to refactor during the error message improvement.

PR #145, opened 2024-10-04
Set centerFocus

evgeniya-aira:
Adds `isCameraFocusCentered` to the `Room` model to get data from the dashboard, indicating whether the agent requested centered focus while taking the picture."


ming-chu (approved):
LGTM, thank you! 😃

PR #149, opened 2024-11-19
Add ringback to the service_request.dart

evgeniya-aira:
### Description 
A new field, `ringback`, has been added to the `service-request` response.

### DIFF
--- a/lib/src/room.dart
+++ b/lib/src/room.dart
@@ -96,7 +96,7 @@ abstract class Room implements Listenable {
   bool get isSensorDataCollectionEnabled;
 
   ///Returns the ringBack mp3 file.
-  String? get ringBackFile;
+  String? get ringbackUrl;
 
   /// Mutes or un-mutes the presentation.
   Future<void> setPresentationMuted(bool muted);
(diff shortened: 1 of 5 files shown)


lib/src/room.dart
```diff
@@ -307,6 +310,9 @@ class KurentoRoom extends ChangeNotifier implements Room {
   bool get isSensorDataCollectionEnabled =>
       _serviceRequest.motionSensorDataCollectionEnabled;
 
+  @override
+  String? get ringBackFile => _serviceRequest.ringbackUrl;
```
bwang-aira, line 314: Is there a reason we don't just call this variable `ringbackUrl`? It's not a file, is it?

evgeniya-aira: It's MP3, but I'll rename it

bwang-aira (approved):

lib/src/room.dart
bwang-aira: I meant, it's a url pointing to a mp3, not the actual mp3 file content, right? Usually we name variables "file" when it's the actual binary file. In this case, it's still a string holding a url. Unless I'm mistaken on the content of the value.

PR #155, opened 2025-01-03
Add new verification constant `kDeviceKeyVerification` 

kasemj:
- Add new verification constant `kDeviceKeyVerification` with value `DEVICE_KEY`


ming-chu (approved):
LGTM

PR #158, opened 2025-01-10
Add service request id to dnc log

HoseinHaqiqian:
This is an informative log, not an exception. We have DNC from agents for this though.


bwang-aira:
If we started a video call without video track, would that imply other issues?

IMO we can leave the error and monitor it as we switch to LiveKit, see if we have any improvements in number of errors.

bwang-aira (approved):

ming-chu:
We don't have the `room.dart` anymore and Kurento room was removed, hence I close it now. @HoseinHaqiqian

PR #160, opened 2025-03-23
Allow a wider range of `flutter_webrtc` dependency

bwang-aira:
Current limitation on `pubspec.yaml` does not allow `flutter_webrtc` to resolve to any higher number. In `explorer-ui` for web, we want to use a later version of `flutter_webrtc`.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
   equatable: ^2.0.5
   flutter:
     sdk: flutter
-  flutter_webrtc: >=0.10.6
+  flutter_webrtc: ">=0.10.6"
   http: ^1.2.0
   intl: ^0.19.0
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -15,15 +15,15 @@ dependencies:
   equatable: ^2.0.5
   flutter:
     sdk: flutter
-  flutter_webrtc: 0.10.6
+  flutter_webrtc: ">=0.10.6"
```
ming-chu, line 18: I think we did that because of next version of `flutter_webrtc` has problem, maybe we should explicit the version.

ming-chu (approved):
LGTM

ming-chu:
We no longer have `flutter_webrtc` in this plugin, hence I close it now.

PR #162, opened 2025-04-25
Expose `intent` in createServiceRequest

kasemj:


lib/src/platform_client.dart
```diff
@@ -284,7 +285,7 @@ class PlatformClient {
       'app': await _appContext,
       'device': await _deviceContext,
       'permissions': {'location': position != null},
-      'intent': 'NONE',
+      'intent': intent ?? 'NONE',
```
ming-chu, line 288: Maybe we could consider keeping it null when it's null instead of introduce 'NONE' on it?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.22.0] - 2025-04-04
+
```
evgeniya-aira, line 8: It looks like  the date wasn't updated

evgeniya-aira (approved):
LGTM

CHANGELOG.md
kasemj: Fixed

lib/src/platform_client.dart
ming-chu: Seems ok since we already use 'NONE' for that before

ming-chu (approved):

lib/src/platform_client.dart
kasemj: But what if the backend is expecting NONE?

ming-chu: agree, backend seems checking it
image

PR #167, opened 2025-05-26
Remove Kurrento, flutter_webrtc and mqtt

kasemj:
Also the following unused plugins.
- meta
- mutex 
- connectivity_plus


example/lib/main.dart
ming-chu, line 1: seems we comment the entire file, maybe consider use /* ... */ for better reading, we will probably remove it eventually though.

ming-chu (approved):
LGTM

PR #172, opened 2025-08-07
Remove try/catch block from _httpRequest

kasemj:
The `SocketException` will be localized in the client to "No internet connection"

### DIFF
--- a/lib/src/platform_client.dart
+++ b/lib/src/platform_client.dart
@@ -1337,38 +1337,33 @@ class PlatformClient {
     Map<String, String>? queryParameters,
     Object? body,
   }) async {
-    try {
-      Uri uri = Uri.https(platformHost, unencodedPath, queryParameters);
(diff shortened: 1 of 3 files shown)


lib/src/platform_client.dart
```diff
@@ -1337,38 +1337,33 @@ class PlatformClient {
...
-          break;
-        case 'PUT':
-          response = await _httpClient.put(uri, headers: headers, body: body);
-          break;
-        default:
-          throw UnsupportedError(method);
-      }
-      return _parseResponse(response.statusCode, response.body);
-    } on SocketException catch (e) {
-      throw PlatformUnknownException(e.message);
```
kasemj, line 1370: This exception wrapping is useless to us.

HoseinHaqiqian: Agreed. Exceptions must be handled within explorer app.

